PR #128, opened 2026-03-06
Added YouTube Connector

dgalluzzo26:
YouTube Connector: PipeLine https://e2-demo-field-eng.cloud.databricks.com/pipelines/c5267988-3c49-4ac1-895b-3c9fb7ae35eb/updates/aca43baf-f8bf-4cea-93b0-7518c1b110b3?o=1444828305810485

LLP: https://databricks.atlassian.net/browse/LPP-5593

Unit Test:
image

E2E Test:

image

image


dgalluzzo26:
@yyoli-db updated rebase and ran a test 
image

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+        if page_token:
+            params["pageToken"] = page_token
+        resp = self._request("GET", "channels", params=params)
+        resp.raise_for_status()
+        data = resp.json()
+        items = data.get("items") or []
+        records = [_flatten_channel(i) for i in items]
+        next_token = data.get("nextPageToken")
+        next_offset = {"pageToken": next_token} if next_token else {"pageToken": None}
+        return iter(records), next_offset
```
yyoli-db: **CRITICAL — re-emits page 1 on the framework's tail call.**

The `LakeflowConnect.read_table` contract (interface/lakeflow_connect.py:92-95) is *equality-based*: the framework calls again with `start_offset = previously-returned end_offset` and stops only when the connector returns the same offset back. Here, when this method previously returned `{"pageToken": None}` at end-of-data, the next call passes that in as `start_offset`. `start_offset.get("pageToken")` returns `None`, the `if page_token:` check on line 354 is falsy, so no `pageToken` is sent — and the API returns **page 1 again**, which gets re-emitted. The connector then returns `{"pageToken": None}` again and the framework converges, but page 1 records have now been emitted twice.

`test_read_terminates` only checks offset convergence, not row uniqueness across calls, so this is invisible in CI.

The same fix `_read_search` / `_read_playlist_items` / `_read_videos` use should be applied here:

```python
def _read_channels(self, start_offset, table_options):
    if start_offset and "pageToken" in start_offset and start_offset.get("pageToken") is None:
        return iter([]), {"pageToken": None}
    ...
```

Same fix needed for `_read_playlists`, `_read_activities`, `_read_comment_threads`, `_read_subscriptions`, `_read_video_categories`.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+        if page_token:
+            params["pageToken"] = page_token
+        resp = self._request("GET", "playlists", params=params)
+        resp.raise_for_status()
+        data = resp.json()
+        items = data.get("items") or []
+        records = [_flatten_playlist(i) for i in items]
+        next_token = data.get("nextPageToken")
+        next_offset = {"pageToken": next_token} if next_token else {"pageToken": None}
+        return iter(records), next_offset
```
yyoli-db: Same pattern as `_read_channels` above — missing the sentinel guard. After the final page returns `{"pageToken": None}`, the framework calls back, this method re-fetches page 1, and emits those rows a second time before converging.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+        if page_token:
+            params["pageToken"] = page_token
+        resp = self._request("GET", "activities", params=params)
+        resp.raise_for_status()
+        data = resp.json()
+        items = data.get("items") or []
+        records = [_flatten_activity(i) for i in items]
+        next_token = data.get("nextPageToken")
+        next_offset = {"pageToken": next_token} if next_token else {"pageToken": None}
+        return iter(records), next_offset
```
yyoli-db: Same duplicate-emission issue as `_read_channels`. Needs the sentinel guard at the top of `_read_activities` to avoid emitting page 1 a second time on the framework's tail call.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+                f"video or channel ({ident}), or it may be restricted. Try a different video_id, "
+                "or see README for channel_id limitations. API reason: " + str(reason)[:200]
+            )
+        resp.raise_for_status()
+        data = resp.json()
+        items = data.get("items") or []
+        records = [_flatten_comment_thread(i) for i in items]
+        next_token = data.get("nextPageToken")
+        next_offset = {"pageToken": next_token} if next_token else {"pageToken": None}
+        return iter(records), next_offset
```
yyoli-db: Same duplicate-emission issue. Also applies to `_read_subscriptions` and `_read_video_categories` below.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+        "snippet_topLevelComment_snippet_authorDisplayName": (
+            top_snip.get("authorDisplayName") if isinstance(top_snip, dict) else None
+        ),
+        "snippet_topLevelComment_snippet_publishedAt": (
+            top_snip.get("publishedAt") if isinstance(top_snip, dict) else None
+        ),
+        "snippet_topLevelComment_snippet_likeCount": (
+            str(top_snip.get("likeCount")) if isinstance(top_snip, dict) and top_snip.get("likeCount") is not None else None
+        ),
+        "snippet_canReply": str(item.get("snippet", {}).get("canReply")) if item.get("snippet", {}).get("canReply") is not None else None,
```
yyoli-db: **Crash on `"snippet": null`.**

`item.get("snippet", {})` only returns the `{}` default when the key is **missing**. If the API returns `"snippet": null` (rare but legal — e.g. some restricted comment threads), this evaluates to `None`, and `.get("canReply")` raises `AttributeError`.

Fix:
```python
snip = item.get("snippet") or {}
...
snip.get("canReply")
```

Same pattern in `_flatten_video_category` (line 224).

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+            headers["Authorization"] = f"Bearer {self._get_access_token()}"
+        backoff = INITIAL_BACKOFF
+        for attempt in range(MAX_RETRIES):
+            resp = self._session.get(url, params=params, headers=headers, timeout=60)
+            if resp.status_code not in RETRIABLE_STATUS_CODES:
+                return resp
+            if attempt < MAX_RETRIES - 1:
+                time.sleep(backoff)
+                backoff *= 2
+        return resp
```
yyoli-db, line 425: Retry loop ignores `Retry-After`. YouTube Data API quota responses include this header; with the current code a quota-exhausted run burns its 5 retries in 1 + 2 + 4 + 8 + 16 = 31 seconds and fails hard, instead of honoring the API's own backoff hint. Also, `_get_access_token` raises a clean `ValueError` on 401, but 400 + `invalid_grant` (revoked refresh token) is the more common failure and falls through to the generic `raise_for_status()` HTTP 400. Worth catching both.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+                "YouTube OAuth returned 401. Check client_id, client_secret, "
+                "and refresh_token; re-run OAuth flow if needed."
+            )
+        resp.raise_for_status()
+        data = resp.json()
+        self._access_token = data["access_token"]
+        self._token_expires_at = time.time() + data.get("expires_in", 3600)
+        return self._access_token
+
+    def _request(self, method: str, path: str, params: dict[str, Any] | None = None) -> requests.Response:
```
yyoli-db: Nit: `method` is ignored — `self._session.get(...)` is hard-coded. Drop the parameter or implement the routing; the docstring already documents it as a GET, so the parameter is dead.

src/databricks/labs/community_connector/sources/youtube/youtube_schemas.py
```diff
@@ -0,0 +1,236 @@
...
+    "search": {
+        "primary_keys": ["search_query", "result_index"],
+        "cursor_field": None,
+        "ingestion_type": "snapshot",
+    },
+    "activities": {
+        "primary_keys": ["id"],
+        "cursor_field": "snippet_publishedAt",
+        "ingestion_type": "snapshot",
+    },
```
yyoli-db, line 218: `activities` is declared `ingestion_type: snapshot` but also carries `cursor_field: snippet_publishedAt`. `_read_activities` never reads or advances this cursor — the field is dead metadata. Either drop the `cursor_field` (snapshot is full-refresh anyway) or wire it through `_read_activities` to filter via `publishedAfter` from the prior offset, and switch the ingestion type accordingly.

src/databricks/labs/community_connector/sources/youtube/connector_spec.yaml
```diff
@@ -0,0 +1,61 @@
...
+      description: >
+        OAuth 2.0 client secret (required if not using api_key).
+
+    - name: refresh_token
+      type: string
+      required: false
+      secret: true
+      description: >
+        Long-lived refresh token with access_type=offline (required if not using api_key).
+
```
yyoli-db, line 55: Nit: all four credentials are marked `required: false`. The UI prompt will surface all four as independently optional. 
you can check other examples that we could in fact have mutiple sub types of a connection so here it will be:
1. api_key
2. client_id, client_secret, refresh_token

all are required

src/databricks/labs/community_connector/sources/youtube/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+requires = ["setuptools>=61.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "lakeflow-community-connectors-youtube"
+version = "0.1.0"
+description = "YouTube connector for Lakeflow Community Connectors"
+requires-python = ">=3.10"
+dependencies = [
+    "pydantic>=2.0.0,<3.0",
```
yyoli-db: Nit: `pydantic>=2.0.0,<3.0` is declared but I don't see `pydantic` imported anywhere in `youtube.py` or `youtube_schemas.py`. Likely inherited from a template — trim if unused, otherwise the connector picks up an unused 5MB dep.

src/databricks/labs/community_connector/sources/youtube/pipeline_spec_search_databricks.json
yyoli-db, line 1: what is this for ?

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/youtube/pipeline_spec_search_databricks.json
dgalluzzo26: removed not needed

src/databricks/labs/community_connector/sources/youtube/pyproject.toml
dgalluzzo26: Removed the dependency

src/databricks/labs/community_connector/sources/youtube/connector_spec.yaml
dgalluzzo26: Restructured into auth_methods:

api_key: requires api_key
oauth: requires client_id, client_secret, refresh_token
Kept top-level connection.oauth for consent flow / u2m. Matches Mixpanel/FHIR pattern.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+            return iter([]), {"pageToken": None}
+        q = (table_options.get("q") or "").strip()
+        if not q:
+            raise ValueError("search requires q (query) in table_options")
+        max_pages_raw = (table_options.get("max_pages") or "").strip()
+        max_pages = int(max_pages_raw) if max_pages_raw.isdigit() else 10
+        max_pages = max(1, min(max_pages, 100))
+        all_records: list[dict] = []
+        page_token: str | None = None
+        batch_id = str(uuid.uuid4())
```
dgalluzzo26: Switched result_index from {uuid}_{position} to str(position) so (search_query, result_index) is stable across snapshot re-runs with SCD_TYPE_1.

Added test_search_result_index_stable_across_runs in test_search_no_duplicate_keys.py.

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Use snip = item.get("snippet") or {} (and top_snip = ... or {}) so null snippets don’t crash. Added test_flatten_null_snippet.py.

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed — added the same sentinel guard as _read_search, _read_playlist_items, and _read_videos at the top of each paginated reader:


if start_offset and "pageToken" in start_offset and start_offset.get("pageToken") is None:
    return iter([]), {"pageToken": None}

dgalluzzo26: Applied to:

_read_channels
_read_playlists
_read_activities
_read_comment_threads
_read_subscriptions
_read_video_categories

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed — added the sentinel guard at the top same as above

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed — added the sentinel guard at the top same as above

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed — added the sentinel guard at the top same as above

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: 1. Retry-After — Added _retry_wait_seconds() so retries use the response Retry-After header when present (seconds), and fall back to exponential backoff otherwise. Quota/throttle responses can wait as long as the API specifies instead of exhausting all 5 attempts in ~31s.

2. invalid_grant on token refresh — _get_access_token now checks for 400 with error == "invalid_grant" and raises a clear ValueError (revoked/expired refresh token — re-run OAuth and update the connection), in addition to the existing 401 handling.

Tests in tests/unit/sources/youtube/test_request_retry_oauth.p

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Good catch — removed the unused method parameter. _request is now:

def _request(self, path: str, params: dict[str, Any] | None = None) -> requests.Response:
All call sites updated (_request("channels", ...) etc.). The implementation remains GET-only, matching the docstring

src/databricks/labs/community_connector/sources/youtube/youtube_schemas.py
dgalluzzo26: Went with Option A — set cursor_field to None for activities

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: @yyoli-db is this fix adequate?

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: @yyoli-db is this fix adequate?

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: @yyoli-db is this fix adequate?

src/databricks/labs/community_connector/sources/youtube/youtube_schemas.py
dgalluzzo26: @yyoli-db is this fix adequate?

src/databricks/labs/community_connector/sources/youtube/connector_spec.yaml
dgalluzzo26: @yyoli-db is this fix adequate?

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: @yyoli-db is this fix adequate?

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: @yyoli-db is this fix adequate?

src/databricks/labs/community_connector/sources/youtube/youtube.py
yyoli-db: **Critical:** `search` PK is non-idempotent across runs.

`result_index = f"{batch_id}_{position}"` with `batch_id = uuid4()` makes the PK fresh every read. With `ingestion_type: snapshot` + `primary_keys = (search_query, result_index)`, the destination grows without bound — every run appends another ~125 rows for the same query, since the SCD_TYPE_1 merge has no PK to match against the prior run. The README's "deduplicate downstream" guidance is a workaround for users; the connector itself is contractually broken for snapshot mode.

Two straightforward fixes:
- **Option A (minimum change):** drop the UUID, use `idx = str(position)`. PK becomes `(q, position)`. Two runs of the same query produce identical PKs; later runs overwrite earlier ones, which is exactly what snapshot mode promises.
- **Option B (natural key):** use the API-provided `id.videoId | id.channelId | id.playlistId` (plus `kind` to disambiguate) as the PK. More work but doesn't depend on result ordering.

The README at line 254 currently labels this as expected behavior — please remove that section once the connector emits stable PKs.

src/databricks/labs/community_connector/sources/youtube/youtube.py
yyoli-db: **Major:** `item.get("snippet", {})` raises `AttributeError` when the API returns `"snippet": null` rather than omitting the key — the `{}` default only applies when the key is *missing*, not when its value is `None`. Restricted comment threads occasionally produce this shape.

Line 175 already uses the safe form `(item.get("snippet") or {})` for the topLevelComment access. Use the same pattern on lines 193, 195–197:
```python
snip = item.get("snippet") or {}
"snippet_canReply": str(snip.get("canReply")) if snip.get("canReply") is not None else None,
"snippet_totalReplyCount": (
    str(snip.get("totalReplyCount")) if snip.get("totalReplyCount") is not None else None
),
```
Or just reuse `_get_nested`.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+    }
+
+
+def _flatten_video_category(item: dict) -> dict:
+    """Flatten a videoCategory resource to schema fields."""
+    return {
+        "id": item.get("id") or "",
+        "snippet_title": _get_nested(item, "snippet.title"),
+        "snippet_assignable": (
+            str(item.get("snippet", {}).get("assignable"))
```
yyoli-db: **Major:** Same `"snippet": null` issue as `_flatten_comment_thread:193`. Replace `item.get("snippet", {}).get(...)` with `(item.get("snippet") or {}).get(...)` or `_get_nested`.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+        headers = {}
+        if not self._api_key:
+            headers["Authorization"] = f"Bearer {self._get_access_token()}"
+        backoff = INITIAL_BACKOFF
+        for attempt in range(MAX_RETRIES):
+            resp = self._session.get(url, params=params, headers=headers, timeout=60)
+            if resp.status_code not in RETRIABLE_STATUS_CODES:
+                return resp
+            if attempt < MAX_RETRIES - 1:
+                time.sleep(backoff)
```
yyoli-db: **Major:** Retry on 429 ignores the `Retry-After` header. YouTube Data API quota responses carry it, often as multi-second to minute-scale waits. Fixed exponential backoff starting at 1s burns through `MAX_RETRIES` in well under a minute when the real quota window is much longer, then fails hard. Honor `Retry-After` on 429 (parse as either seconds or HTTP-date) and fall back to `backoff` only if the header is missing.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+                "grant_type": "refresh_token",
+            },
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
+        )
+        if resp.status_code == 401:
+            raise ValueError(
+                "YouTube OAuth returned 401. Check client_id, client_secret, "
+                "and refresh_token; re-run OAuth flow if needed."
+            )
+        resp.raise_for_status()
```
yyoli-db, line 402: `_get_access_token` handles 401 cleanly but `raise_for_status()` covers `invalid_grant` (which Google returns as **400**, not 401) with a generic `HTTPError`. Revoked / expired refresh tokens are a common case — surface a clear `ValueError` for `status=400` + `error=invalid_grant` (parse the response body) the same way 401 is handled. Otherwise users see an opaque `400 Client Error` and have no idea their refresh token needs re-issuing.

src/databricks/labs/community_connector/sources/youtube/youtube_schemas.py
```diff
@@ -0,0 +1,236 @@
...
+    "search": {
+        "primary_keys": ["search_query", "result_index"],
+        "cursor_field": None,
+        "ingestion_type": "snapshot",
+    },
+    "activities": {
+        "primary_keys": ["id"],
+        "cursor_field": "snippet_publishedAt",
```
yyoli-db: **Major:** `activities` declares `ingestion_type: snapshot` AND `cursor_field: snippet_publishedAt`. `_read_activities` never reads `start_offset` for a cursor or advances one — the field is dead and misleads anyone reading the metadata. Either:
- drop `cursor_field` and accept full-refresh snapshot semantics, or
- switch `ingestion_type` to `incremental` and wire the cursor through `_read_activities` via the `publishedAfter` API parameter.

The second option would also stop activities from re-reading the entire channel history on every run.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+            records = [_flatten_video(i) for i in items]
+            return iter(records), {"pageToken": None}
+        if use_chart:
+            max_pages_raw = (table_options.get("max_pages") or "").strip()
+            max_pages = int(max_pages_raw) if max_pages_raw.isdigit() else 20
+            max_pages = max(1, min(max_pages, 100))
+            all_records: list[dict] = []
+            page_token: str | None = None
+            for _ in range(max_pages):
+                p = {"part": "snippet,statistics,contentDetails", "maxResults": _max_results(table_options, cap=50), "chart": "mostPopular"}
```
yyoli-db: **Minor:** `cap=50` is hard-coded at each call site (lines 343, 368, 409, 429, 447, 510, 571, 594) and `cap=100` at line 537 for comment_threads. A future caller can silently lower a cap by typo, and the comment_threads-specific value (100) drifts apart from everywhere else. Consider a constant table:
```python
_MAX_RESULTS_CAP_BY_ENDPOINT = {"comment_threads": 100, "_default": 50}
```
or a per-endpoint constant. Not blocking, but reduces footgun risk.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,607 @@
...
+        page_token = start_offset.get("pageToken")
+        if page_token:
+            params["pageToken"] = page_token
+        resp = self._request("GET", "commentThreads", params=params)
+        if resp.status_code == 403:
+            try:
+                err = resp.json().get("error", {})
+                reason = err.get("errors", [{}])[0].get("reason", "") or err.get("message", "")
+            except Exception:
+                reason = resp.text or "unknown"
```
yyoli-db: **Minor:** The 403 reason extraction caps via `str(reason)[:200]` on the JSON-parse path (line 558), but on the fallback path `reason = resp.text or "unknown"` — `resp.text` can be many KB of HTML/JSON on edge errors and is interpolated unbounded into the `ValueError` message. Add the same slice on `resp.text`: `reason = (resp.text or "unknown")[:200]`.

src/databricks/labs/community_connector/sources/youtube/connector_spec.yaml
```diff
@@ -0,0 +1,61 @@
...
+# Connection Parameters
+# Provide EITHER api_key (for public data) OR OAuth (client_id, client_secret,
+# refresh_token) for private/mine data. API key is simpler; OAuth needed for
+# mine=true, subscriptions, etc.
+# ============================================================================
+connection:
+  parameters:
+    - name: api_key
+      type: string
+      required: false
```
yyoli-db: **Minor (#10 + #14 combined):** All four credentials are marked `required: false`, with the XOR (api_key OR all-three-OAuth) only enforced in `__init__` of the Python class. The UI surfaces these YAML descriptions verbatim, so users see four optional fields with no hint of the XOR contract. Add the XOR explanation explicitly in each `description` (or in a header comment users can see), so the UI prompt is self-explanatory.

Separately, `channel_id` in `external_options_allowlist` (line 61) means different things depending on the table — for `_read_subscriptions` it means "channelId whose subscribers I want," which the README at line 133 notes is 403-on-public-channels in most cases. Flagging this in the YAML option description (not just the README) would warn users at config time rather than at first run.

src/databricks/labs/community_connector/sources/youtube/pipeline_spec_search_databricks.json
```diff
@@ -0,0 +1,82 @@
...
+        "table_configuration": {
+          "mine": "true"
+        }
+      }
+    },
+    {
+      "table": {
+        "source_table": "comment_threads",
+        "table_configuration": {
+          "video_id": "PASTE_VIDEO_ID_FROM_YOUR_CHANNEL"
```
yyoli-db: **Info:** Hard-coded placeholder `"video_id": "PASTE_VIDEO_ID_FROM_YOUR_CHANNEL"`. Acceptable for a sample file, but please add a top-level comment in the JSON (or rename the file to make its sample-only nature explicit) so users don't accidentally run it as-is.

src/databricks/labs/community_connector/sources/youtube/pyproject.toml
yyoli-db: **Info:** `pydantic>=2.0.0,<3.0` is declared but I don't see `pydantic` imported anywhere in the connector source. Likely inherited from a template — please drop it unless you intend to use it, so the connector's dependency footprint stays minimal.

src/databricks/labs/community_connector/sources/youtube/youtube.py
yyoli-db: I think the problem still stays. However, I just realize that the `channel` and all other tables have ingestion_type = snapshot, so these incremental logic is not really needed. We need to re-read the table fully for every update.

Some tables cannot be incrementally ingested most likely (as Youtube does not provide you an API to filter out records using things like timestamp). So:

- can we confirm what tables can be incrementally read what cannot be?
- for the ones that cannot be incrementally read (it is possible none of the tables can), get rid of the logic of dealing with offset

src/databricks/labs/community_connector/sources/youtube/youtube.py
yyoli-db: see comments above.

src/databricks/labs/community_connector/sources/youtube/youtube_schemas.py
yyoli-db: If there is no cursor field, we can remove it from the map. No need to specifiy None

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: agreed — all nine tables are ingestion_type: snapshot, so framework pageToken offset is batching within a trigger, not incremental ingestion. None of the YouTube list endpoints support true CDC-style reads;

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: see comments above

src/databricks/labs/community_connector/sources/youtube/youtube_schemas.py
dgalluzzo26: Removed

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Addressed in bca597a: dropped UUID batch_id; result_index is str(position) (0-based). PK (search_query, result_index) is stable across snapshot re-runs for the same query. Covered by test_search_result_index_stable_across_runs.

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed in bca597a: _flatten_comment_thread and _flatten_video_category use snip = item.get("snippet") or {} and guard canReply / totalReplyCount / assignable when null. Covered by test_flatten_null_snippet.py.

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed in bca597a — _flatten_video_category uses snip = item.get("snippet") or {} for snippet_assignable. Tested in test_flatten_video_category_snippet_null

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed in bca597a: _request sleeps _retry_wait_seconds(resp, backoff) on 429/5xx, honoring numeric Retry-After with exponential backoff fallback. Covered by test_request_retry_oauth.py.

src/databricks/labs/community_connector/sources/youtube/youtube.py
dgalluzzo26: Fixed in bca597a: _get_access_token parses 400 responses for invalid_grant and raises a clear ValueError with re-auth guidance before raise_for_status(). Covered by test_get_access_token_invalid_grant_raises_value_error.

dgalluzzo26:
@yyoli-db 
<h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 590; margin-top: 16px; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">hread 1: Snapshot / framework offset (<code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">pageToken</code><span> </span>across calls)</h3><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">@yyoli-db — agreed. All nine tables are<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">ingestion_type: snapshot</code><span> </span>and none support true incremental CDC via framework offset. YouTube list APIs only give<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">pageToken</code><span> </span>for<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">within-request</span><span> </span>pagination, not a durable checkpoint across pipeline runs.<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">published_after</code><span> </span>on<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">search</code><span> </span>/<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">activities</code><span> </span>is a<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">query filter</span><span> </span>only.</p><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">Addressed in<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">e37371f</code>:</span></p><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;">Every table now<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">drains all API pages inside a single<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">read_table</code><span> </span>call</span><span> </span>(<code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">_drain_paginated_list</code><span> </span>/ internal loops).</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;">Framework offset is<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">{"done": True}</code></span><span> </span>(actitime snapshot pattern) — no<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">pageToken</code><span> </span>in connector offsets.</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;">Tail calls with<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">{"done": True}</code><span> </span>short-circuit to empty with<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">no extra API requests</span><span> </span>(fixes the original tail-call duplicate-page-1 class of bugs).</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;">Docs updated (<code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgb(60, 124, 171); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word; cursor: pointer;"><span class="md-inline-path-filename">README.md</span></code>,<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgb(60, 124, 171); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word; cursor: pointer;"><span class="md-inline-path-filename">youtube_api_doc.md</span></code>).</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;">Tests:<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgb(60, 124, 171); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word; cursor: pointer;"><span class="md-inline-path-filename">test_pagination_multi_page.py</span></code>,<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgb(60, 124, 171); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word; cursor: pointer;"><span class="md-inline-path-filename">test_pagination_tail_call.py</span></code><span> </span>(all 9 tables).</li></ul><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">Incremental feasibility (all tables today):</span></p><div class="ui-scroll-area" data-scroll-padding="4" data-visibility="hover" data-direction="horizontal" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; --scrollbar-size: 6px; --scrollbar-inset: 0px; --scrollbar-thumb-top-offset: 6px; --scroll-area-scroll-padding: 4px; position: relative; overflow: hidden; display: grid; grid-template: 1fr / 1fr; margin: 1em 0px; border-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); border-style: solid; border-width: 0.833333px; border-image: none 100% / 1 / 0 stretch; border-radius: 6px; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div class="ui-scroll-area__viewport" style="grid-area: 1 / 1; min-height: 0px; max-height: 100%; border-radius: inherit; overflow: auto hidden; scroll-padding: 0px 4px; overscroll-behavior: contain auto; scrollbar-width: none !important;"><div class="ui-scroll-area__content" style="box-sizing: border-box; width: auto; min-width: 100%; max-width: 100%; min-height: 100%; display: flex; flex-direction: row; height: fit-content;">
Table | Framework incremental? | Notes
-- | -- | --
All 9 | No | Snapshot full re-read per run; optional published_after is a filter, not a cursor

</div></div></div><hr class="my-6 border-border" data-streamdown="horizontal-rule" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 16px 0px; border-right: none; border-bottom: none; border-left: none; border-image: initial; border-top: 0.833333px solid color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 590; margin-top: 16px; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Thread 2:<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">activities</code><span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">cursor_field</code><span> </span>vs snapshot</h3><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">@yyoli-db — took<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">Option A</span><span> </span>(snapshot full-refresh).<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">_read_activities</code><span> </span>never advanced a framework cursor.</p><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">History:</span></p><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;"><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">bca597a</code>: removed misleading<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">snippet_publishedAt</code><span> </span>cursor (<code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">cursor_field: None</code>)</li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;"><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">e37371f</code>: removed<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">cursor_field</code><span> </span>from<span> </span><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">all</span><span> </span>snapshot metadata maps entirely (not<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">None</code>) — matches actitime</li></ul><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">activities</code><span> </span>metadata is now only<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">primary_keys</code><span> </span>+<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">ingestion_type: snapshot</code>.<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">published_after</code><span> </span>remains an optional table option to narrow the API query.</p><hr class="my-6 border-border" data-streamdown="horizontal-rule" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 16px 0px; border-right: none; border-bottom: none; border-left: none; border-image: initial; border-top: 0.833333px solid color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><h3 class="mt-6 mb-2 font-semibold text-xl" data-streamdown="heading-3" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin-bottom: 4px; font-size: 1.214em; line-height: 1.42; --tw-font-weight: 600; font-weight: 590; margin-top: 16px; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Earlier review items (already in<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">bca597a</code><span> </span>/<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">3b690ef</code>, unchanged in<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">e37371f</code>)</h3><p style="--tw-space-y-reverse: 0; margin-block: 0px 16px; margin: 6px 0px; word-break: break-word; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">For reference — these were addressed in prior commits; threads can be resolved:</p><ul class="list-inside list-disc whitespace-normal [li_&amp;]:pl-6" data-streamdown="unordered-list" style="--tw-space-y-reverse: 0; margin-block: 0px 16px; list-style-type: disc; white-space: normal; padding-left: 2em; display: flex !important; flex-direction: column !important; gap: 6px !important; margin: 6px 0px !important; color: rgba(20, 20, 20, 0.92); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(252, 252, 252); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">Search PK</span>:<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">result_index = str(position)</code><span> </span>—<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">test_search_result_index_stable_across_runs</code></li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">Null<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">snippet</code></span>:<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">_flatten_comment_thread</code><span> </span>/<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">_flatten_video_category</code><span> </span>use<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">item.get("snippet") or {}</code><span> </span>—<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgb(60, 124, 171); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word; cursor: pointer;"><span class="md-inline-path-filename">test_flatten_null_snippet.py</span></code></li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;">Retry-After on 429</span>:<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">_retry_wait_seconds</code><span> </span>—<span> </span><code class="md-clickable-code md-inline-path-filename-like" role="button" tabindex="0" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgb(60, 124, 171); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word; cursor: pointer;"><span class="md-inline-path-filename">test_request_retry_oauth.py</span></code></li><li class="py-1 [&amp;&gt;p]:inline" data-streamdown="list-item" style="padding-block: 4px; word-break: break-word; padding: 0px !important;"><span class="font-semibold" data-streamdown="strong" style="--tw-font-weight: 600; font-weight: 600;"><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">invalid_grant</code></span>: clear<span> </span><code class="" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">ValueError</code><span> </span>in<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">_get_access_token</code><span> </span>—<span> </span><code class="md-inline-variable-like" style="padding: 1.5px 4px; border-radius: 5px; background-color: color(srgb 0.0784314 0.0784314 0.0784314 / 0.0737255); color: rgba(20, 20, 20, 0.92); font-family: Menlo, Monaco, &quot;Courier New&quot;, monospace; font-size: 0.9em; line-height: 1.4; word-break: normal; overflow-wrap: break-word;">test_get_access_token_invalid_grant_raises_value_error</code></li></ul>

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,712 @@
...
+            return done
+        video_ids = (table_options.get("video_ids") or "").strip()
+        use_chart = (table_options.get("chart") or "").lower() == "mostpopular"
+        if video_ids:
+            params = {
+                "part": "snippet,statistics,contentDetails",
+                "id": video_ids,
+                "maxResults": _max_results(table_options, cap=50),
+            }
+            resp = self._request("videos", params=params)
```
yyoli-db: Silent truncation past 50 IDs. The `video_ids` branch issues a single `videos.list` request with no pagination, so passing >50 comma-separated IDs returns only the first 50 and silently drops the rest (the API caps `id=`-filtered responses at `maxResults` <= 50). Consider chunking `video_ids` into batches of 50 and looping, or at minimum documenting the 50-ID ceiling next to the `videos` option in the README

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,712 @@
...
+    """
+
+    def __init__(self, options: dict[str, str]) -> None:
+        super().__init__(options)
+        self._api_key = (options.get("api_key") or "").strip()
+        self._client_id = (options.get("client_id") or "").strip()
+        self._client_secret = (options.get("client_secret") or "").strip()
+        self._refresh_token = (options.get("refresh_token") or "").strip()
+        has_oauth = bool(self._client_id and self._client_secret and self._refresh_token)
+        if self._api_key and has_oauth:
```
yyoli-db, line 346: MINOR (optional cleanup): the `else` branch below re-checks the OAuth triple even though `has_oauth` already captured it. The logic is correct (it raises a clear error on partial-OAuth credentials), just structurally redundant. Could simplify to a single `if not self._api_key and not has_oauth: raise`. Not a defect.

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,712 @@
...
+    raw = (table_options.get("max_pages") or "").strip()
+    if not raw:
+        return default
+    try:
+        return max(1, min(int(raw), cap))
+    except ValueError:
+        return default
+
+
+def _snapshot_done_short_circuit(
```
yyoli-db: we do not need this. For all the snapshot read, we could simply ignore the `start_offset` and emit `None` for the function's returned offset.

yyoli-db (changes requested):

dgalluzzo26:
@yyoli-db latest updates are in `91598a9` (rebased onto current `youtube`):
- Switched snapshot termination offsets to `None` across all YouTube tables (removed done-sentinel short-circuit helper).
- Fixed `videos` `video_ids` truncation by batching comma-separated IDs in chunks of 50 per API request.
- Added `timeout=60` for OAuth token refresh POST in `_get_access_token`.
- Aligned simulator corpus for `search.list` with live response shape (`regionCode`) to remove validation drift.
- Updated docs/tests/generated single-file source accordingly.
Validation/tests run:
- Pylint (CI-equivalent flags) passes for YouTube source files.
- Simulate-mode suite: `40 passed, 2 skipped`.
- Live-mode suite (API-key config): `39 passed, 3 skipped`.
- Live validation artifact is now clean:
  - `endpoints_validated=9`
  - `endpoints_mismatched=0`
  - `endpoints_missing_spec=0`

dgalluzzo26:
image


image

yyoli-db (approved):

dgalluzzo26:
Reran test and added OAuth.

image


image

dgalluzzo26:
@yyoli-db Thanks for the review — addressed both items below.

CI: generated merged source
Regenerated _generated_youtube_python_source.py after the latest youtube.py changes and rebased onto current master. The generate-merged-sources check should pass on this branch.

Standard OAuth (U2M) — aligned with Gmail #220
Rebased onto master (including Gmail U2M OAuth) and applied the same pattern to YouTube:

connector_spec.yaml: OAuth auth method is now client_id + client_secret only (no manual refresh_token in the connection form). Added connection.oauth with flow: u2m, Google endpoints/scopes, and extra_auth_params (access_type=offline, prompt=consent).
youtube.py: Supports UC-injected access_token (opaque, no in-connector refresh) for u2m / u2m_per_user, plus legacy refresh_token + client_id + client_secret for older/static connections. api_key unchanged for public data.
Docs/tests: README and youtube_api_doc.md updated; auth unit tests cover access_token and refresh-token paths.
PKCE note: Initial U2M create on e2-demo failed with CONNECTION_COMMUNITY_OAUTH_U2M must include pkce_verifier when pkce: false (matching Gmail’s spec). Set pkce: true in connector_spec.yaml (45e0d61) — U2M connection creation then succeeded.

Validated locally:

Simulate tests: 44 passed
Live dev_config.json: API key + OAuth refresh + injected access_token paths verified against YouTube API
Workspace: dgalluzzo_youtube created as OAUTH_U2M on e2-demo; pipeline retest successful
Happy to squash commits on merge if preferred. Please re-run CI when convenient (safe-to-test).

src/databricks/labs/community_connector/sources/youtube/youtube.py
```diff
@@ -0,0 +1,706 @@
...
+            timeout=60,
+        )
+        if resp.status_code == 401:
+            raise ValueError(
+                "YouTube OAuth returned 401. Check client_id, client_secret, "
+                "and refresh_token; re-run OAuth flow if needed."
+            )
+        if resp.status_code == 400:
+            try:
+                if resp.json().get("error") == "invalid_grant":
```
yyoli-db, line 392: nit: non-JSON 400 body raises the wrong (less actionable) error. `requests`' `JSONDecodeError` subclasses `ValueError`, so if the token endpoint returns a non-JSON 400 body, `resp.json()` here raises `JSONDecodeError`, which the `except ValueError: raise` below catches and re-raises — bypassing the intended fall-through to the clean `resp.raise_for_status()`. Caller still errors, just with a raw `JSONDecodeError` instead of an actionable HTTP message.

Suggest parsing first, then branching:
```python
try:
    payload = resp.json()
except Exception:
    payload = {}
if payload.get("error") == "invalid_grant":
    raise ValueError("... invalid_grant ...")
```

Edge case only (Google normally returns JSON on 400) and only reachable on the legacy refresh-token path, not the preferred UC-injected token.

yyoli-db (approved):

PR #155, opened 2026-04-23
Add Palantir Foundry ontology connector

zaibos:
Connector for ingesting data from Palantir Foundry ontology object types into Databricks via the Lakeflow Community Connector framework.

## Features

- **Dynamic schema discovery** from Palantir object type definitions. Unknown/unmapped property types fall back to `StringType` with a warning; `__`-prefixed system fields (`__rid`, `__primaryKey`, `__apiName`, `__title`) are excluded from the discovered schema (documented).
- **Two ingestion modes** — Snapshot (SCD_TYPE_1) and incremental CDC (SCD_TYPE_2), selected by the presence of `cursor_field`.
- **`loadObjects` with `snapshot=true`** for consistent, multi-page pagination (`nextPageToken` threading).
- **Composite CDC cursor.** Incremental runs resume on a composite `(cursor_field, tiebreaker)` cursor — server-side `orderBy [cursor, tiebreaker]` + a `where` clause of `cursor > prev OR (cursor == prev AND tiebreaker > prev)`. This makes the ordering total, so rows that share a non-unique cursor value (e.g. many rows at the same timestamp) are **never skipped or duplicated** across a batch boundary. The tiebreaker defaults to the ontology's declared primary-key property and is overridable via `tiebreaker_field`. (System fields like `__primaryKey`/`__rid` can't be the cursor/tiebreaker — Foundry rejects them in `orderBy`/`where`.)
- **Max-cursor lookup** via the `search` endpoint (`orderBy desc, pageSize=1`) feeding an early-exit short-circuit that skips the `loadObjects` read only when the dataset's max cursor is strictly *below* the checkpoint.
- **`_init_time` admission cap** bounds each trigger for timestamp cursors (Trigger.AvailableNow termination).
- **Memory-efficient snapshot** streaming via generators (one page at a time, no OOM).
- **Bounded retry** with jittered exponential backoff + `Retry-After` on 429/503/network errors — applied on both the data path and the schema-discovery GET. Non-transient 4xx/5xx fail fast and surface Palantir's structured error body (`errorCode`/`errorName`).
- **HTTPS enforced** — `http://` hostnames are rejected so the bearer token is never sent in cleartext.

## Files

- `palantir.py` — connector implementation (LakeflowConnect interface)
- `_generated_palantir_python_source.py` — merged deployable file
- `connector_spec.yaml` — connection parameters + `external_options_allowlist`
- `README.md` — user-facing documentation
- `palantir_api_doc.md` — Palantir Foundry API reference
- `source_simulator/specs/palantir/` — simulator spec (`endpoints.yaml`) + corpus
- Test suite using the shared `LakeflowConnectTests` harness

## Testing

- Full unit suite passes in default (simulate) mode, including a dedicated regression for the composite-cursor tied-value case, schema edge cases, retry behavior, and multi-page pagination.
- Validated **live** against a real Foundry tenant via record mode: the connector drained a 475K-row object type incrementally (composite cursor converges) plus a snapshot table, with **0 spec drift** across all endpoints.


yyoli-db:
Hi, @zaibos 
we have made bunch of changes recently so that we could re-enable the git actions/workflow (new policy and requirements due to the supply chain security issue in the whole company). 

Please check the latest master and run some of these skills and agents. 
**In particular**, there is no a `/self-review-connector` skill you need to run which will help to catch issues early on.

tests/unit/sources/palantir/configs/dev_table_config.json
yyoli-db, line 1: Test gap. In LakeflowConnectTests, _opts(table) returns table_configs.get(table, {}). The committed                                                 
  tests/unit/sources/palantir/configs/dev_table_config.json keys options under "FlightsFinal", but the simulator corpus (object_types_list.json) only
  exposes "ExampleFlight". So in simulate mode every shared test runs ExampleFlight with {}, which means ingestion_type is always "snapshot" in CI. The  
  CDC path (_read_incremental, _get_max_cursor_via_aggregate, where: gt building) is never exercised by the shared suite — only by the bespoke
  test_search_endpoint_coverage and the two mocked unit tests. test_read_terminates, test_read_table, etc. all run snapshot.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,650 @@
...
+            pk_fields = primary_key
+        else:
+            pk_fields = ["__primaryKey"]
+
+        # Build metadata based on whether cursor field is specified
+        if cursor_field:
+            # Incremental (CDC) mode
+            metadata = {
+                "primary_keys": pk_fields,
+                "cursor_field": cursor_field,
```
yyoli-db, line 390: qq: does Palantir have some sort of shared cursor_field like `lastModifiedTimestamp` across all the tables? or it is like a database, where the tables are very self-defined?

tests/unit/sources/palantir/configs/dev_table_config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "FlightsFinal": {
```
yyoli-db, line 2: this table does not seem appearing in your testing simulation corpus

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,650 @@
...
+            )
+        )
+        prev_max_cursor = start_offset.get("max_cursor_value") if start_offset else None
+
+        # Step 1: Get current max cursor via aggregate (lightweight).
+        current_max_cursor = self._get_max_cursor_value(table_name, cursor_field)
+
+        # Step 2: Cap at _init_time so this trigger run doesn't chase
+        # data arriving after connector start.
+        if current_max_cursor and current_max_cursor > self._init_time:
```
yyoli-db: the cursor comes from user defined column, which might not be timestamp? maybe have some check before hand on the cursor times?

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,650 @@
...
+        self._session.headers.update({
+            "Authorization": f"Bearer {self.token}",
+            "Content-Type": "application/json",
+        })
+
+        # Admission control: cap cursor at init time so a single trigger
+        # run only drains data that existed when the connector started.
+        self._init_time = datetime.now(timezone.utc).strftime(
+            "%Y-%m-%dT%H:%M:%SZ"
+        )
```
yyoli-db: we have to make sure this is in the same format and timezone setting as the one returned by Palantir APIs. Right now, it does not seem so.
The comparison will be messed up otherwise

tests/unit/sources/palantir/configs/dev_table_config.json
zaibos: - 676f0da - dev_table_config.json now keys options under both ExampleFlight (the simulator corpus's table) and FlightsFinal (the live API's), with arrivalTimestamp as the unified cursor field. This fires _read_incremental and _get_max_cursor_via_aggregate end-to-end through the shared suite in simulate mode (ingestion_type: 'cdc' verified locally; aggregate POST is hit during test_read_table and test_read_terminates).

- 9f2bf61 - The where-clause branch can't be reached by test_read_terminates even with the ExampleFlight config: the simulator's corpus/aggregate.json returns a static value, so on the second read prev_max == current_max and _read_incremental early-returns at palantir.py:526-528 before reaching the clause builder. The dedicated test forces the branch with a deliberately stale start_offset so prev < current, and asserts on the resulting object_set shape passed to _fetch_page. Happy to make aggregate advance via a custom handler if you'd prefer that over a focused unit test

src/databricks/labs/community_connector/sources/palantir/palantir.py
zaibos: Palantir Foundry does not give every dataset or ontology table a single built-in column like lastModifiedTimestamp that works the same everywhere. It behaves much more like many self-defined schemas (plus platform metadata around datasets), not one universal relational catalog with standard columns on every table.

tests/unit/sources/palantir/configs/dev_table_config.json
zaibos: 882fafe - Added both ExampleFlight and FlightsFinal test. Both tables now exercise CDC in simulate mode

src/databricks/labs/community_connector/sources/palantir/palantir.py
zaibos: Fixed in 73594f3 + 65b0a2c. The cap now parses both _init_time and the cursor into tz-aware UTC datetimes (handles Z, +00:00, -05:00, naive, fractional, date-only) so format/timezone differences don't break the comparison. Non-string / unparseable cursors return None from the parser and the cap is skipped — termination still bounded by the offset-unchanged check. max(prev, current) is also gated on type  equality, falling back to the current value rather than crashing on mismatch. Coverage in               TestPalantirCursorTypes.

src/databricks/labs/community_connector/sources/palantir/palantir.py
zaibos: Fixed in 65b0a2c. Both _init_time and the cursor value are now parsed into tz-aware UTC datetime objects via _to_utc_datetime before comparing — handles Z, explicit offsets like +00:00 / -05:00, naive (assumed UTC), fractional seconds, and date-only forms. Lex compare on raw strings would miss cases like "2026-05-09T08:00:00-05:00" (== 13:00 UTC) since - < Z lexically; the parsed comparison gets it right. Cap value still preserves the cursor's shape (date stays date, full timestamp stays full timestamp) via length-matched slicing of _init_time. New test test_cap_compares_across_timezone_offsets locks in the timezone case.

zaibos:
# Self-Review — palantir

**Overall: 100 / 100 — READY**

Run at: 2026-06-10T08:37:21Z

> Audited at commit `16dcfb9` (round-3 review fixes: dropped unused
> `pydantic` dep, added two type hints, corrected the `max_tiebreak_value`
> offset doc). Pylint 10.00/10, 68 passed / 2 skipped (simulate).
> **Fresh live run against `16dcfb9`** drained ~475K rows in 3m20s with
> **0 drift** (3/3 endpoints validated, 0 mismatched) — the previously
> stale B11 artifact is now re-stamped newer than the source. Every
> section passes; no findings.

## Top recommendations

_None blocking._

## A. Connector implementation — 25 / 25
- ✅ A1. Class extends `LakeflowConnect` — `palantir.py:32`
- ✅ A2. All abstract methods present — `palantir.py:178,198,352,404`
- ✅ A3. No `cdc_with_deletes` table (`ingestion_type` only `cdc`/`snapshot` — `palantir.py:391,397`) → `read_table_deletes` not required
- ✅ A4. Read pattern: server-side composite `(cursor, tiebreaker)` (Strategy B) incremental + generator snapshot — deliberate
- ✅ A5. `_init_dt`/`_init_time` cap present and applied per-record — `palantir.py:106-107,923,939`
- ✅ A6. `max_records_per_batch` honored on incremental (`palantir.py:846`); snapshot intentionally exempt (`palantir.py:475,491`)
- ✅ A7. All 3 HTTP calls carry `timeout=` — GET 30s (`:155`), search peek 30s (`:689`), loadObjects 120s (`:1061`)
- ✅ A8. No `IntegerType`/`MapType`
- ✅ A9. Nested types preserved via recursive `StructType`/`ArrayType`
- ✅ A10. Imports clean (stdlib, `requests`, `pyspark`, interface only) — `palantir.py:8-27`
- ✅ A11. Pylint clean — 10.00/10
- ✅ A12. No experimental ingestion-agent surface

- ✅ B1. Test file exists
- ✅ B2. Subclasses `LakeflowConnectTests` — `test_palantir_lakeflow_connect.py:13`
- ✅ B3. `connector_class` set — `:14`
- ✅ B4. `simulator_source = "palantir"` — `:15`
- ✅ B5. `replay_config` keys (token, hostname, ontology_api_name) match `__init__` — `:16-20` ≡ `palantir.py:60,64,68`
- ✅ B6. `endpoints.yaml` present
- ✅ B7. Corpus present (`object_types_list.json`, `objects.json`)
- ✅ B8. All 3 connector URLs (objectTypes/loadObjects/search) have spec entries
- ✅ B9. `test_read_terminates` passes (simulate)
- ✅ B10. All tests pass — 68 passed / 2 skipped
- ✅ B11. **Fresh live run, 0 drift** — captured against `16dcfb9` (3/3 endpoints validated, 0 mismatched); ~475K-row FlightsFinal incremental drain in 3m20s. Artifact now newer than source/spec.
- ✅ B12. Coverage full (3 of 3 spec endpoints hit)
- ✅ B13. `synthesize_future_records` present on cursor endpoint — `endpoints.yaml:43`
- ✅ B14. No write-back surface (N/A)

## C. Artifacts — 28 / 28
- ✅ C1. Compiles (`py_compile` clean)
- ✅ C2. API doc present, documents composite cursor + offset shape
- ✅ C3. `connector_spec.yaml` parses; connection params + `external_options_allowlist`
- ✅ C4. README present, mentions tables + all 3 params
- ✅ C5. `pyproject.toml` parses; `requests` in deps
- ✅ C6. Generated source newer than `palantir.py` (regenerated at commit)
- ✅ C8. Corpus files valid JSON
- ✅ C9. Test class exists
- ✅ C10. `pyspark` is dev-only — not in `[project.dependencies]`

## D. Security — 22 / 22
- ✅ D1. No hardcoded secrets (only match is the `token:` docstring line — `palantir.py:49`)
- ✅ D2. No `eval`/`exec`/`compile`
- ✅ D3. No `subprocess`/`os.system`/`shell=True`
- ✅ D4. No `verify=False`
- ✅ D5. No `pickle.load`
- ✅ D6. No unsafe `yaml.load`
- ✅ D7. No path traversal
- ✅ D8. No secrets in logs (retry breadcrumb logs backoff + attempt only)
- ✅ D9. No `http://` base URLs in production (only the `http://` rejection guard) — `palantir.py:79-85`
- ✅ D10. All dependencies bounded (`>=X,<Y`)

- ✅ E1. Tables dynamically discovered (`list_tables` — `palantir.py:178`); example tables documented in README + API doc
- ✅ E2. Connection params agree: `__init__` ≡ spec ≡ README (token, hostname, ontology_api_name)
- ✅ E3. Per-table schemas: discovered properties documented in API doc
- ✅ E4. Primary keys consistent (declared PK / `__primaryKey` fallback)
- ✅ E5. `external_options_allowlist` covers all 4 `table_options.get` keys (cursor_field, page_size, max_records_per_batch, tiebreaker_field)
- ✅ E6. README quick-start uses spec connection-parameter keys
- ✅ E7. README consistent with API doc on auth/pagination

tests/unit/sources/palantir/configs/.gitignore
yyoli-db, line 1: let's delete this

tests/unit/sources/palantir/configs/dev_config.json.example
yyoli-db, line 1: let's delete

tests/unit/sources/palantir/README.md
yyoli-db, line 1: let's delete

tests/unit/sources/palantir/configs/.gitignore
zaibos: removed

tests/unit/sources/palantir/configs/dev_config.json.example
zaibos: removed

tests/unit/sources/palantir/README.md
zaibos: removed

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+        endpoint first as a "lightweight max-cursor lookup" and fell
+        back to ``search``. In practice Palantir returns 500 for
+        aggregate against the ontology object types we use, so the
+        fallback path was the only one that ever produced a result.
+        We removed the aggregate code path to skip the doomed HTTP
+        round-trip on every CDC tick. ``search`` (``orderBy desc,
+        limit 1``) works universally and the payload-size difference
+        vs aggregate is negligible (one record vs one metric).
+        """
+        return self._get_max_cursor_via_search(table_name, cursor_field)
```
yyoli-db, line 667: **Major (M1) — dead code + doc/spec drift.** After the Strategy B refactor (commit `59cabb6`), `_read_incremental` drives the `where: gt` filter directly from `start_offset.get("max_cursor_value")` and never calls `_get_max_cursor_value`. This helper (and `_get_max_cursor_via_search` below) are reachable only from the test suite — see `test_search_endpoint_coverage` and the entire `TestPalantirMaxCursorLookup` class, which exist purely to keep the path alive.

Meanwhile the README still advertises the search pre-compute as a headline feature:
- `README.md:12` — *"Pre-Computed Cursor Lookup: Single API call ... to determine max cursor value for CDC checkpointing"*
- `README.md:126-127` — *"Pre-computes max cursor value via the search endpoint"*
- `palantir_api_doc.md` and `endpoints.yaml` advertise `search` as a connector endpoint.

Please pick one:
1. Re-add the early-exit short-circuit (mirroring the example connector's `since >= self._init_ts` skip): fetch `new_max` via `_get_max_cursor_via_search` once, and if `new_max <= prev_max_cursor`, return `iter([]), start_offset` without hitting `loadObjects`. This is also what the previous reviewer's feedback assumed was happening.
2. Or delete the helpers, drop `test_search_endpoint_coverage` + `TestPalantirMaxCursorLookup`, prune the `search` endpoint from `endpoints.yaml`, and rewrite the README/api-doc sections to describe what the connector actually does now (server-side `where: gt`, no pre-flight).

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+        max_retries = 5
+        for attempt in range(max_retries):
+            try:
+                response = self._session.post(url, json=body, timeout=120)
+                if response.status_code in (429, 503):
+                    wait = 2 ** attempt
+                    time.sleep(wait)
+                    continue
+                response.raise_for_status()
+            except requests.RequestException as e:
```
yyoli-db: **Major (M2) — non-transient HTTP errors are retried.** `Response.raise_for_status()` raises `HTTPError` (a `RequestException` subclass) for every 4xx/5xx, so the `except requests.RequestException` arm catches **everything** including 400, 401, 403, 404. With `max_retries=5`, a real auth failure (expired token) or misconfiguration (wrong ontology name) burns `1+2+4+8 = 15 s` of `time.sleep` and 5 wasted load-balancer hits before the user sees the error.

Suggested shape:
```python
if response.status_code in (429, 503):
    time.sleep(2 ** attempt); continue
response.raise_for_status()  # let non-transient 4xx/5xx propagate immediately
...
except (requests.ConnectionError, requests.Timeout) as e:
    if attempt < max_retries - 1:
        time.sleep(2 ** attempt); continue
    raise RuntimeError(...) from e
```
That way only connection/timeout errors plus 429/503 are retried; everything else fails fast.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+        # run only drains data that existed when the connector started.
+        self._init_time = datetime.now(timezone.utc).strftime(
+            "%Y-%m-%dT%H:%M:%SZ"
+        )
+        self._default_max_records_per_batch = 100_000
+
+        # Initialize caches for performance
+        self._schema_cache: Dict[str, StructType] = {}
+        self._metadata_cache: Dict[str, dict] = {}
+        self._object_types_cache: Dict[str, dict] = None
```
yyoli-db: **Minor (m6) — type annotation lies about the runtime value.** Field is annotated `Dict[str, dict]` but initialised to `None`. Use `Optional[Dict[str, dict]]` (and add `Optional` to the `typing` import on line 11) so mypy/IDE inference stays honest.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+        self._session.headers.update({
+            "Authorization": f"Bearer {self.token}",
+            "Content-Type": "application/json",
+        })
+
+        # Admission control: cap cursor at init time so a single trigger
+        # run only drains data that existed when the connector started.
+        self._init_time = datetime.now(timezone.utc).strftime(
+            "%Y-%m-%dT%H:%M:%SZ"
+        )
```
yyoli-db: **Minor (m10) — re-parsing `_init_time` once per record.** This stores the cap as a string and `_to_utc_datetime` re-parses it on every iteration of `_read_incremental` (line 555, called once per record). Cheaper:
```python
self._init_dt = datetime.now(timezone.utc)
self._init_time = self._init_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
```
and use `self._init_dt` directly in the cap loop. Drops 100 K redundant ISO parses per microbatch at the default cap.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+    Palantir Foundry connector implementing the LakeflowConnect interface.
+
+    Supports:
+    - Dynamic schema discovery from Palantir object type definitions
+    - Both snapshot and incremental (CDC) ingestion modes
+    - Cursor-based pagination using nextPageToken
+    - User-configurable cursor fields for incremental sync
+    """
+
+    def __init__(self, options: Dict[str, str]) -> None:
```
yyoli-db, line 43: **Minor (m4)** — no `super().__init__(options)`. `LakeflowConnect.__init__` sets `self.options = options` as the documented base convention (see `interface/lakeflow_connect.py:21` and `sources/example/example.py:26`). Functionally harmless today because the framework keeps its own copy of options, but breaks the moment a future base-class helper reads `self.options`.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+        url = f"{self.base_url}/api/v2/ontologies/{self.ontology_api_name}/objectTypes"
+
+        try:
+            response = self._session.get(url, timeout=30)
+            response.raise_for_status()
+        except requests.RequestException as e:
+            raise Exception(
+                f"Failed to list object types from Palantir ontology "
+                f"'{self.ontology_api_name}': {str(e)}"
+            )
```
yyoli-db, line 169: **Minor (m5)** — `raise Exception(...)` is too broad (pylint `broad-exception-raised`). The example connector uses `RuntimeError` for the same shape (`sources/example/example.py:68`); please switch here and at `palantir.py:672, 686`. Bonus: `raise RuntimeError(...) from e` preserves the chain.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+                raise Exception(
+                    f"Failed to fetch data after {max_retries} retries: {str(e)}"
+                )
+
+            data = response.json()
+            records = data.get("data", [])
+            next_page_token = data.get("nextPageToken")
+
+            # Small delay for rate limiting
+            time.sleep(0.1)
```
yyoli-db: would the retry path already takes care of the rate limiting to some degree? 
unconditional sleep seems less desirable, especially 0.1 second for every iteration

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,688 @@
...
+            "integer": LongType(),
+            "long": LongType(),
+            "double": DoubleType(),
+            "float": DoubleType(),
+            "boolean": BooleanType(),
+            "timestamp": StringType(),  # Keep as ISO 8601 string
+            "date": StringType(),  # Keep as ISO date string
+            "datetime": StringType(),
+            "attachment": StringType(),  # JSON representation
+            "decimal": DoubleType(),  # Simplified for MVP
```
yyoli-db: **Minor (m7)** — `decimal → DoubleType` is documented as "Simplified for MVP" but isn't called out under README **Limitations**. For finance/measurement ontologies this is a silent precision-loss risk; please add a one-liner under **Limitations** in `README.md:344-355`.

Can't we directly use DecimalType of Spark? was it due to precision limitation?

src/databricks/labs/community_connector/sources/palantir/ingest_notebook.py
```diff
@@ -0,0 +1,72 @@
...
+# MAGIC %md
+# MAGIC ## Step 2: Define the pipeline specification
+# MAGIC
+# MAGIC Edit the values below to match your environment:
+# MAGIC - `connection_name`: Your Unity Catalog connection name
+# MAGIC - `source_table`: Palantir object type API name (e.g., "ExampleFlight")
+# MAGIC - `destination_catalog/schema/table`: Where to land the data in Databricks
+# MAGIC - `scd_type`: SCD_TYPE_1 (upsert), SCD_TYPE_2 (history), or APPEND_ONLY
+# MAGIC - `cursor_field`: Set for incremental sync, omit for snapshot (full refresh)
+# MAGIC - `page_size`: Records per API call (default 100, max 1000)
```
yyoli-db: **Minor (m3)** — comment says `page_size` default is 100 and max 1000. The actual default is 1000 and the cap is 10000 (`palantir.py:388, 650`; README agrees). Please align with the code so users don't undersize their pages by 10x.

yyoli-db (changes requested):

yyoli-db:
please check http://go/community-connectors and file the LPP like here: http://go/community-connectors/tracking

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,773 @@
...
+        max_records = int(
+            table_options.get(
+                "max_records_per_batch", self._default_max_records_per_batch
+            )
+        )
+
+        if start_offset:
+            next_offset = start_offset
+        else:
+            next_offset = {"done": "true"}
```
yyoli-db: **Minor (m1, repeat from prior round) — `_read_snapshot` returns a non-empty offset on the first call, causing the framework to re-paginate the whole dataset on the termination check.**

The base interface contract is equality-only ("Pagination stops when the returned offset equals start_offset"). Returning `{"done": "true"}` on call #1 means the framework calls `read_table` a second time with `start_offset = {"done": "true"}`. The harness's `_validate_termination` drains the iterator every call (`test_suite.py:687`), so on call #2 the connector re-enters `_generate_all_pages`, re-issues `_fetch_page` round-trips for every page, and only then returns `{"done":"true"}` again. For a 10 M-row table that's a full re-pagination per snapshot trigger.

The reference connector returns `{}` on every snapshot call (`example.py:194`); the framework treats equal-empty offsets as terminated on the first iteration. Simplest fix: drop the `if start_offset` branch and always return `{}`:

```python
return self._generate_all_pages(
    table_name, page_size, max_records=max_records
), {}
```

This was raised as `m1` in the prior round and isn't addressed in commit `42e0618` despite the M2–M10 sweep.

src/databricks/labs/community_connector/sources/palantir/pipeline_spec.example.py
```diff
@@ -0,0 +1,342 @@
...
+
+cursor_field (optional):
+    - Field name for incremental sync
+    - Typically a timestamp field (e.g., "updatedAt", "departureTimestamp")
+    - If not specified, uses snapshot mode (full refresh)
+
+page_size (optional):
+    - Number of records per API request
+    - Default: 100
+    - Maximum: 1000
```
yyoli-db: **Minor (doc-drift) — `page_size` configuration reference is stale.**

This block says:

```
- Default: 100
- Maximum: 1000
```

The code default is `1000` and the cap is `10000` (`palantir.py:406, 580, 723`). README (line 107) and `ingest_notebook.py` (line 40) already agree with the code; this reference at the bottom of the example file is the last copy still on the stale numbers from the prior round (sibling of the `m3` finding that was fixed in `ingest_notebook.py`).

src/databricks/labs/community_connector/sources/palantir/ingest_notebook.py
yyoli-db, line 1: please just remove this file. we don't have to merge this. the main ingest.py will be provided by the UI flow as well as the cli flow

src/databricks/labs/community_connector/sources/palantir/pipeline_spec.example.py
yyoli-db, line 1: let's do not create such .py file. if these examples are really necessary, put them as part of the connector doc
`palantir_api_doc`

src/databricks/labs/community_connector/sources/palantir/palantir.py
zaibos: resolved. Made changes as suggested. commit id - 6047b3b

src/databricks/labs/community_connector/sources/palantir/pipeline_spec.example.py
zaibos: These examples are not needed, I have taken backup of these files into another feature branch and deleted them from this feature.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,773 @@
...
+        # sleep and 5 wasted load-balancer hits before the user sees
+        # the actual error.
+        max_retries = 5
+        last_status: int = -1
+        for attempt in range(max_retries):
+            try:
+                response = self._session.post(url, json=body, timeout=120)
+            except (requests.ConnectionError, requests.Timeout) as e:
+                if attempt < max_retries - 1:
+                    time.sleep(2 ** attempt)
```
yyoli-db: **P1 — Retry backoff has no jitter.**

`time.sleep(2 ** attempt)` synchronises every concurrent Spark task to the same wake-up cadence (1s, 2s, 4s, 8s). Against Palantir's 5000 RPM / 30-concurrent-connection per-user limit, a 429 storm caused by a wide pipeline would have every task retry in lockstep, which prolongs the storm. Add randomised jitter, e.g. `time.sleep(2 ** attempt * (0.5 + random.random()))`. Also consider honouring a `Retry-After` header if Palantir sets one — the current path discards it and the user only sees the count + last status.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,773 @@
...
+
+        # Build base URL
+        self.base_url = f"https://{self.hostname}"
+
+        # Configure authenticated session
+        self._session = requests.Session()
+        self._session.headers.update({
+            "Authorization": f"Bearer {self.token}",
+            "Content-Type": "application/json",
+        })
```
yyoli-db, line 99: `requests.Session` is not closed. `self._session = requests.Session()` is created in `__init__` with no `close()`. Spark's data source lifecycle can construct many `PalantirLakeflowConnect` instances per query; each holds an open connection pool. Add a `close()` (or `__del__`) that calls `self._session.close()`. Not a correctness bug, but will leak ports on long-running pipelines.

tools/scripts/merge_exclude_config.json
```diff
@@ -17,12 +17,18 @@
     ],
     "google_analytics_aggregated": [
       "pipeline_spec.example.py"
+    ],
+    "palantir": [
+      "pipeline_spec.example.py",
+      "ingest_notebook.py"
     ]
```
yyoli-db, line 20: you don't need these anymore after the removal of the 2 files

tests/unit/sources/palantir/configs/dev_table_config.json
```diff
@@ -0,0 +1,7 @@
...
+        "cursor_field": "arrivalTimestamp",
+        "page_size": "100",
+        "max_records_per_batch": "50"
+    }
+}
```
yyoli-db, line 10: test gap:

Every table here carries `cursor_field`, so the shared suite's `test_read_table` / `test_read_terminates` only exercises CDC. Please add at least one snapshot-mode entry (either a second table without `cursor_field`, or `FlightsFinal` without it) so the snapshot read path is covered by CI. With the corpus and `synthesize_future_records: 3` setup, a snapshot table with a small `max_records_per_batch` would have caught the data-loss bug flagged on `palantir.py:420`.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,773 @@
...
+        self.hostname = options.get("hostname")
+        if not self.hostname:
+            raise ValueError("Missing required parameter 'hostname'")
+
+        self.ontology_api_name = options.get("ontology_api_name")
+        if not self.ontology_api_name:
+            raise ValueError("Missing required parameter 'ontology_api_name'")
+
+        # Build base URL
+        self.base_url = f"https://{self.hostname}"
```
yyoli-db: nit: what if user specificies `https://` in the `self.hostname`?

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,842 @@
...
+        # emitted = 0
+        while True:
+            records, next_page_token = self._fetch_page(
+                object_set, page_token, page_size, order_by_field
+            )
+            for record in records:  # pylint: disable=use-yield-from
+                # if max_records and emitted >= max_records:
+                #     return
+                yield record
+                # emitted += 1
```
yyoli-db: **P2 (new) — dead commented-out code from the cap-removal rework.**

`# emitted = 0` (398), `# if max_records and emitted >= max_records: return / # emitted += 1` (404-405, 407) are useful only as breadcrumbs for the prior review. Once that context is gone they become noise. Strip them.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,842 @@
...
+            return self._read_incremental(
+                table_name, start_offset, table_options, cursor_field
+            )
+        else:
+            return self._read_snapshot(table_name, start_offset, table_options)
+
+    def _generate_all_pages(
+        self, table_name: str, page_size: int,
+        where_clause: dict = None,
+        max_records: int = 0,
```
yyoli-db: **P2 (new) — `max_records: int = 0` is now an unused parameter.**

After the snapshot fix, neither caller passes `max_records`: snapshot mode omits it; incremental mode caps the records list after iteration in its own loop, not by passing `max_records` to the generator. The parameter and the corresponding docstring sentence (lines 390-394) are dead — drop them so future readers don't think the cap is happening inside `_generate_all_pages`.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,842 @@
...
+            }
+
+        # Retry policy: only transient failures retry. Everything else
+        # (4xx/5xx that isn't 429 or 503) propagates immediately —
+        # otherwise a real auth/config error like 401 (expired token)
+        # or 404 (wrong ontology name) would burn 1+2+4+8 = 15s of
+        # sleep and 5 wasted load-balancer hits before the user sees
+        # the actual error.
+        max_retries = 5
+        last_status: int = -1
```
yyoli-db: **P2 (new) — `last_status` is dead.**

Assigned at line 792 and updated at line 810, but never read. The only consumer was the commented-out post-loop raise (lines 839-842, also worth deleting). Remove both the declaration and the reassignment — the live code path doesn't need it.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,842 @@
...
+
+            data = response.json()
+            records = data.get("data", [])
+            next_page_token = data.get("nextPageToken")
+            return records, next_page_token
+
+        # raise RuntimeError(
+        #     f"Failed to fetch data after {max_retries} retries: "
+        #     f"last status {last_status}"
+        # )
```
yyoli-db: same. Please do not have this dead code commented out. Just remove them

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,842 @@
...
+        # where the next call can resume via ``where: gt
+        # last_emitted_cursor``. Snapshot mode runs in a single
+        # framework-driven pass (``apply_changes_from_snapshot``) with
+        # no mid-snapshot checkpoint, so capping here would silently
+        # drop records past the cap — a data-loss bug.
+
+        # if start_offset:
+        #     next_offset = start_offset
+        # else:
+        #     next_offset = {"done": "true"}
```
yyoli-db: what is this for? should we just remove this

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,1080 @@
...
+        ``"...-05:00"`` < ``"...Z"`` lexically even when the actual
+        moments compare the other way).
+        """
+        if not isinstance(value, str):
+            return None
+        normalised = (
+            value[:-1] + "+00:00" if value.endswith("Z") else value
+        )
+        try:
+            parsed = datetime.fromisoformat(normalised)
```
yyoli-db, line 523: **MINOR: `datetime.fromisoformat` space-separator parsing is 3.11+, but `pyproject.toml` declares `requires-python = ">=3.10"`.**

The live corpus shows `arrivalTimestamp` as `"2026-04-19 06:20:35.746137"` (space separator, no `T`/`Z`). On **Python 3.10**, `fromisoformat` only accepts strict `datetime.isoformat()` output and raises `ValueError` on the space form (the broadened parser landed in 3.11). The `try/except → return None` here means a 3.10 run degrades silently: timestamp cursors fall back to lexicographic compare in `_cursor_strictly_greater` and the `_init_dt` admission cap disengages (it's gated on `cursor_dt is not None`).

Not a crash, and DBR runtimes are 3.11+, but it's inconsistent with the declared 3.10 floor. Suggest either bumping `requires-python` to `>=3.11`, or normalizing the separator (`value.replace(" ", "T", 1)`) before parsing. A test asserting the space form parses would also lock this in — `test_to_utc_datetime_parses_known_palantir_formats` currently only covers `T` forms.

.gitignore
```diff
@@ -68,3 +68,9 @@ tests/unit/sources/*/cassettes/*.json
 # Self-review reports (output of the self-review-connector skill). Local
 # checkpoint for a developer; not committed by default.
 tests/unit/sources/*/SELF_REVIEW.md
+
+# Pipeline-deployment notebooks auto-created by the Databricks SDP UI in
+# git-linked workspace repos. Per-pipeline artifacts, not part of the
+# connector library. Connector code lives only under
+# src/databricks/labs/community_connector/sources/<name>/.
+/src/ingest.py
```
yyoli-db: let's just remove this. The solution is that we should not sync this file from the databricks workspace repo to your local and commit it and push.

yyoli-db (approved):

src/databricks/labs/community_connector/sources/palantir/palantir.py
zaibos: Thanks — good catch on the test gap, fixed. But I checked the parsing premise on a real **Python 3.10.20** interpreter and the space-separated form actually parses fine there:

```
>>> datetime.fromisoformat("2026-04-19 06:20:35.746137")
datetime.datetime(2026, 4, 19, 6, 20, 35, 746137)
```

`fromisoformat` has accepted the **space** separator since 3.7 — it reads the char at index 10 as the separator and doesn't require `T`. What 3.11 broadened was the bare `Z` suffix, arbitrary fractional-digit counts, and missing separators — not the space. And the one form 3.10 genuinely rejects (trailing `Z`) is already normalized to `+00:00` *before* `fromisoformat` is called. So both Palantir formats (`arrivalTimestamp`'s space form and the `T…Z` forms) parse on 3.10, the `_init_dt` cap stays engaged, and there's no silent lexicographic fallback.

So no `requires-python` bump or `replace(" ", "T", 1)` is needed — the first would tighten the floor for a non-existent incompatibility, and the second is a functional no-op that'd misleadingly imply the space form is unsupported.

I did add your suggested test, though — `test_to_utc_datetime_parses_known_palantir_formats` now asserts the space form (and a space+`Z` variant) in addition to the `T` forms. It passes on 3.10.20 and on the default interpreter, locking the behavior in. (commit 7297299)

.gitignore
zaibos: Removed (commit db7d9ea). Agreed — gitignoring `/src/ingest.py` was treating the symptom; the real fix is to not sync that SDP-workspace artifact into the local clone in the first place. The file isn't tracked and doesn't exist in the repo, so dropping the rule (and its comment) is clean with no side effects.

src/databricks/labs/community_connector/sources/palantir/palantir.py
```diff
@@ -0,0 +1,1080 @@
...
+                    f"boundary. Set 'tiebreaker_field' to a non-nullable "
+                    f"unique property, or run this table in snapshot mode."
+                )
+            records.append(record)
+            last_emitted_cursor = cursor_value
+            last_emitted_tiebreak = tiebreak_value
+
+        # No records emitted → terminate the microbatch with the
+        # same offset so Spark sees "no progress" and stops.
+        if not records:
```
yyoli-db, line 972: NIT (cosmetic, no change required): on a true first run that emits zero records (`start_offset` falsy, `prev_max_cursor is None`), this returns `{"max_cursor_value": None}`. Harmless — it's JSON-serializable and the next call converges by equality (`test_read_terminates` passes) — but `{"max_cursor_value": None}` vs `{}` is a slightly awkward offset to checkpoint. Flagging for awareness only.

PR #176, opened 2026-05-11
feat(actitime): add actiTIME community connector

### LINKED ISSUE #178: actitime: `_read_time_window` silently drops overflow when window record count exceeds `stopAfter`
## Summary

In `src/databricks/labs/community_connector/sources/actitime/actitime.py`, the `_read_time_window` method (used by the `timetrack` and `leavetime` append tables) issues a single GET per sliding window with `stopAfter` as a server-side result cap, then unconditionally advances the cursor to `window_end_date`.

If a single window genuinely contains more rows than `stopAfter` (default `1000`), the overflow records are dropped silently because the cursor moves past them on the next microbatch.

Reference:
- `actitime.py:662-746` (`_read_time_window`)
- `actitime.py:698` — `stop_after = int(table_options.get("stopAfter", "1000"))`
- `actitime.py:744-746` — cursor advances unconditionally to `window_end_date`

## Why this matters

For low-volume tenants this never trips. For higher-volume tenants ingesting `timetrack`, the default `window_days=7` × `stopAfter=1000` cap means any week with >1000 time entries silently loses data — there is no warning, log line, or retry.

This is a different bug class than the offset/limit pagination fix in #176 (commit \`bb61503\`), which addressed early termination on offset/limit list endpoints. The `/timetrack` and `/leavetime` endpoints do not expose offset/limit — only \`dateFrom\`/\`dateTo\`/\`stopAfter\` — so the safe mitigation is window shrinking, not pagination.

## Proposed fix

When \`len(data) >= stop_after\` after a window read, treat the window as oversized and either:

1. **Shrink the window** — re-read \`[read_from, midpoint]\` and \`[midpoint+1, window_end_date]\` and recurse, halving until each sub-window's record count is below \`stopAfter\`; only then advance the cursor.
2. **Surface an error** — fail fast with a clear message asking the user to lower \`window_days\` or raise \`stopAfter\`.

Option 1 is more user-friendly; option 2 is safer if we don't want to fan-out API calls unbounded.

## Acceptance criteria

- A window read that hits the \`stopAfter\` cap does not silently drop records.
- A test in \`tests/unit/sources/actitime/\` covers the overflow case using simulator-synthesized records to drive a window past the cap.

## Found in

PR #176 self-review. Filed as a follow-up so the PR remains scoped to @yyoli-db's specific feedback (offset/limit pagination + README/code drift).

fnapolitano73:
## What's included

- **API research doc** — `src/databricks/labs/community_connector/sources/actitime/actitime_api_doc.md` (16 tables across cdc / append / snapshot)
- **Implementation** — `src/databricks/labs/community_connector/sources/actitime/actitime.py` (HTTP Basic auth, offset/limit pagination for list endpoints, per-user `userIds` fan-out for `/timetrack` and `/leavetime`, `_init_date` termination cap)
- **Simulator spec + corpus** — `source_simulator/specs/actitime/` (16 endpoints, 12 corpus files; 4 endpoints intentionally have no corpus because they're tenant-feature-gated)
- **Connector spec** — `connector_spec.yaml` (3 connection params: `base_url`, `username`, `password`; 9 external option allowlist entries including `user_batch_size`)
- **README** — `src/databricks/labs/community_connector/sources/actitime/README.md`
- **Tests** — `tests/unit/sources/actitime/test_actitime_lakeflow_connect.py` + `__init__.py` (includes #178 fan-out coverage tests)
- **Per-connector packaging** — `pyproject.toml`
- **Merged source** — `_generated_actitime_python_source.py` (regenerated after every connector edit)

## Tables (16)

| Type | Tables |
|---|---|
| cdc | customers, projects, tasks |
| snapshot | users, departments, userGroups, userRates, typesOfWork, leaveTypes, workflowStatuses, timeZoneGroups, settings, holidays |
| append | timetrack, leavetime, approvalStatus |

`users` was modelled as `snapshot` (rather than `cdc`) because actiTIME's `/users` response carries no timestamp field. CDC tables are full-refresh + upsert-on-id since actiTIME has no `modifiedAt` filter.

## Review-driven changes folded into this PR

This PR was previously split into a stacked follow-up for the larger fix (#180), but per @yyoli-db's request the changes are now consolidated here so review stays in one place. Closes #178.

| Commit | What it addresses |
|---|---|
| `bb61503` | First review pass — `_read_offset_limit` premature termination when server caps page size; README/code drift (users `cdc`→`snapshot`; `timetrack`/`leavetime` composite PKs; schema notes cleanup) |
| `3264ec7` | Issue #178 — `_read_time_window` per-user `userIds` fan-out (default `user_batch_size=100`) replacing the silent `stopAfter=1000` truncation. Demotes `stopAfter` to an opt-in passthrough. Adds two tests for fan-out coverage + zero-users cursor advancement. |

## Test results

- **Simulate mode**: 8 passed / 2 skipped (skips are `test_partition_requires_lakeflow_connect` — no partition mixin — and `test_read_table_deletes` — no `cdc_with_deletes` tables)
- **Live mode** (against `online.actitime.com/<tenant>`): 8 passed / 2 skipped, same skip rationale
- **Spec drift validation** (live validation report next to the cassette): `endpoints_total=16, endpoints_validated=12, endpoints_mismatched=4, endpoints_missing_spec=0`

### About the 4 spec mismatches

All four are identical: `status_code: live=404 spec=200` for `/approvalStatus`, `/holidays`, `/settings`, `/userGroups`. These endpoints are **tenant-feature-gated** on the actiTIME tenant used for record-mode tests — actiTIME returns 404 with `{key, message}` when the feature is disabled. The connector handles this correctly: `_get_json_or_none` treats 404 as "table absent on this tenant" and emits zero rows with a `{"done": True}` cursor for clean trigger termination.

This isn't a connector defect — the endpoint spec schema (`endpoint_spec.ResponseShape`) currently can't express "200 OR 404 acceptable" for tenant-conditional features. Worth a separate framework follow-up to add multi-status spec entries.

## Self-review

```
Overall: 99 / 100 — READY
```

Full report posted as a sticky PR comment via `/self-review-connector for actitime`. Highlights:

- A. Connector implementation — 23 / 23 (pylint 10.00/10)
- B. Testing & simulator — 34 / 35 (1-point deduction for the framework limitation above)
- C. Artifacts — 25 / 25
- D. Security — 22 / 22
- E. Cross-doc consistency — 11 / 11

## Live deployment

Validated end-to-end on the `e2-demo-field-eng` workspace:

- UC connection: `actitime_dev` (BASIC_AUTH via `GENERIC_LAKEFLOW_CONNECT`)
- Ingestion pipeline: `actitime_pipeline_v2` (serverless, 16 tables → `fna_demo.bronze_actitime_v2`)
- Deployed with `community-connector create_pipeline ... --use-local-source` (local code overlaid on the repo clone).

## Reviewer notes

- Built using `/develop-connector actitime` → `/validate-connector actitime` → `/self-review-connector for actitime` as the documented Phase 1 / Phase 2 / pre-PR flow.
- All shared framework code (`libs/`, `pipeline/`, `interface/`, `source_simulator/`) is untouched. Only files under `sources/actitime/`, `source_simulator/specs/actitime/`, and `tests/unit/sources/actitime/` are modified.
- Five commits, intentionally kept small for review:
  1. `f4ad4ee` — initial Phase 1 build via `/develop-connector`
  2. `8882717` — Phase 2 record-mode drift fixes (envelope unwrap, 404 tolerance, schema corrections from live cassettes)
  3. `0073c20` — pre-PR self-review polish (pylint clean, drop undocumented aliases, fix `/userRates/{user_id}` wrapper)
  4. `bb61503` — first-pass review feedback (offset/limit pagination + README/code drift)
  5. `3264ec7` — issue #178 (per-user fan-out for `/timetrack` and `/leavetime`)

Closes #178.


fnapolitano73:
# Self-Review — actitime

**Overall: 99 / 100 — READY**

Run at: 2026-05-12 18:31 UTC (fifth pass; record-mode re-validated against
live tenant after #178 was folded into PR #176 per @yyoli-db's request.
HEAD is now `3264ec7` — includes both `bb61503` review feedback and the
per-user `userIds` fan-out fix that previously lived in the now-closed
PR #180).

No BLOCKER failures. No MAJOR findings. The only remaining observation
is the previously-documented MINOR framework-spec limitation for
tenant-feature-gated 404s.

## Top recommendations

1. **MINOR (carried over) — B11/E framework limitation.** 4 endpoints
   (`/approvalStatus`, `/holidays`, `/settings`, `/userGroups`) return
   HTTP 404 on the test tenant because those features are disabled.
   The connector handles this correctly via `_get_json_or_none`; the
   validator flags them because `endpoint_spec.ResponseShape` cannot
   declare multiple acceptable statuses or mark an endpoint as
   "tenant-feature-gated". Not a connector defect — worth filing a
   framework follow-up.
2. **Resolved in this PR — #178 silent overflow.** Previously tracked
   as out-of-scope on the prior self-review pass. Commit `3264ec7`
   replaces the single-GET-with-`stopAfter` shape in `_read_time_window`
   with per-user `userIds` fan-out (default `user_batch_size=100`).
   `stopAfter` is demoted to an opt-in passthrough. Two new tests cover
   the fan-out path (≥ 2 distinct user_ids, no composite-PK dupes) and
   the zero-users cursor-advancement edge case.

## A. Connector implementation — 23 / 23

- ✅ A1. Class extends `LakeflowConnect` — `actitime.py:356`
- ✅ A2. All abstract methods implemented — `actitime.py:505, 509, 513, 518`
- ✅ A3. No `cdc_with_deletes` tables, so `read_table_deletes` correctly absent
- ✅ A4. Read pattern deliberate: snapshot via `_read_snapshot`; per-user `userIds`
       fan-out + `dateFrom`/`dateTo` for `timetrack`/`leavetime` (post-#178);
       date-window only for `approvalStatus`; offset/limit via `_read_offset_limit`
       for paginated lists — `actitime.py:519-532`
- ✅ A5. `_init_date` cap present — `actitime.py:389`; cursor short-circuit guards in
       all 3 append-table readers (`actitime.py:689, 766`)
- ✅ A6. `max_records_per_batch` honored — `actitime.py:816`
- ✅ A7. Single retry-loop call site at `actitime.py:410` passes
       `timeout=DEFAULT_TIMEOUT`
- ✅ A8. Schemas use `LongType`, `StructType`/`ArrayType` — no `IntegerType` or `MapType`
- ✅ A9. `get_table_schema` returns nested structures correctly
       (e.g. `userRates.leave_rates` is `ArrayType(StructType(...))`)
- ✅ A10. Imports clean — stdlib + `requests` + `pyspark.sql.types` +
        `databricks.labs.community_connector.interface` — `actitime.py:56-76`
- ✅ A11. **Pylint: 10.00 / 10** (CI flags)

## B. Testing & simulator validation — 34 / 35

- ✅ B1. Test file exists — `tests/unit/sources/actitime/test_actitime_lakeflow_connect.py`
- ✅ B2. Subclasses `LakeflowConnectTests`
- ✅ B3. `connector_class = ActitimeLakeflowConnect`
- ✅ B4. `simulator_source = "actitime"`
- ✅ B5. `replay_config` covers `__init__` keys (`base_url`, `username`, `password`)
- ✅ B6. `endpoints.yaml` present — 16 endpoints
- ✅ B7. Corpus has 12 JSON files (4 tenant-feature-gated tables intentionally
       have no corpus)
- ✅ B8. All 16 spec paths match the URLs the connector constructs via `_url(path)`
- ✅ B9. `test_read_terminates` passes in default mode
- ✅ B10. 8 passed / 2 skipped in default (simulate) mode (existing 6 + 2 new #178 tests)
- ⚠️ B11. **Fresh.** Validation report at
       `tests/unit/sources/actitime/cassettes/TestActitimeConnector.json.validation.json`
       regenerated 2026-05-12 18:31 UTC against the live actiTIME tenant — newer than
       the consolidated `actitime.py` HEAD `3264ec7`. Summary:
       `endpoints_total=16, endpoints_validated=12, endpoints_mismatched=4,
       endpoints_missing_spec=0`. The 4 mismatches are the same documented
       tenant-feature-gated 404s — see Top recommendations #1. Marked `warn` (not
       `fail`) because the connector handles these correctly; the divergence is the
       spec schema, not the code.
- ✅ B12. Coverage: 28 API calls recorded across the live run; all 16 spec
       endpoints were exercised (`cassettes/TestActitimeConnector.json.coverage.json`)
- ✅ B13. `synthesize_future_records: {cursor_field: …}` declared on 3 cursor-bearing
       append tables — `endpoints.yaml:279, 300, 315`
- ✅ B14. Write-back not implemented; `test_utils_class` correctly absent

## C. Artifacts — 25 / 25

- ✅ C1. `actitime.py` exists; `py_compile` clean
- ✅ C2. `actitime_api_doc.md` exists; all 16 tables present under "Object Schema"
- ✅ C3. `connector_spec.yaml` parses; has `connection.parameters` and
       `external_options_allowlist`
- ✅ C4. `README.md` exists; mentions every table and every connection parameter
- ✅ C5. `pyproject.toml` parses; `dependencies` includes `requests>=2.28.0,<3.0`,
       `pyspark`, `pydantic`, `lakeflow-community-connectors`
- ✅ C6. `_generated_actitime_python_source.py` regenerated after the source edits
       (mtime parity with `actitime.py`)
- ✅ C7. `endpoints.yaml` parses with `endpoint_spec.load_specs`
- ✅ C8. All 12 corpus JSON files are valid JSON
- ✅ C9. Test class present (covered by B1)

## D. Security — 22 / 22

- ✅ D1. No hardcoded secrets — only `password` parameter read from connection options
- ✅ D2. No `eval` / `exec` / `compile` of dynamic strings
- ✅ D4. No `verify=False` on `requests` calls
- ✅ D5. No `pickle.load(s)` on untrusted data
- ✅ D6. No `yaml.load` — connector does not parse YAML at runtime
- ✅ D7. No path-traversal sinks from `table_options`
- ✅ D8. No secrets in logs — `auth` header constructed but never logged
- ✅ D9. No `http://` base URLs in production code
- ✅ D10. Dependencies pinned with upper bounds in `pyproject.toml`

- ✅ E1. `list_tables()` (16 tables) ≡ API doc "Object Schema" ≡ README "Tables"
- ✅ E2. Connection-parameter keys aligned across `__init__`, `connector_spec.yaml`,
       and README: `base_url`, `username`, `password`
- ✅ E3. Schema column names match API doc fields (post-snake-case mapping)
- ✅ E4. Primary keys consistent with API doc; **README corrected** on commit
       `bb61503` so `users` is `snapshot` and `timetrack`/`leavetime` carry their
       composite PKs `(user_id, date, task_id)` and `(user_id, date, leave_type_id)`
- ✅ E5. `external_options_allowlist` covers every `table_options.get(…)` /
       `table_options[…]` access in the source: `start_date`, `window_days`,
       `lookback_days`, `stopAfter`, `dateFrom`, `dateTo`, `limit`,
       `max_records_per_batch`
- ✅ E6. README quick-start uses spec param names
- ✅ E7. No detected contradictions between README and API doc

---

## Score detail

| Section | Points | Max |
|---|---|---|
| A. Implementation | 23 | 23 |
| B. Testing & simulator | 34 | 35 |
| C. Artifacts | 25 | 25 |
| D. Security | 22 | 22 |
| E. Cross-doc consistency | 11 | 11 |
| **Total** | **115** | **116** |
| **Normalized** | **99** | **100** |

Verdict: **READY** (90+ band). Sole remaining finding is the
framework-spec limitation around tenant-feature-gated 404s — not a
connector defect.

src/databricks/labs/community_connector/sources/actitime/actitime.py
```diff
@@ -0,0 +1,1058 @@
...
+    # CDC tables. ``customers``, ``projects`` and ``tasks`` expose
+    # a ``created`` epoch-ms timestamp on the API; we keep it as a proxy
+    # cursor but still perform a full-list refresh + upsert on id since
+    # there is no ``modifiedAt`` filter.
+    "customers": {"primary_keys": ["id"], "cursor_field": "created", "ingestion_type": "cdc"},
+    "projects": {"primary_keys": ["id"], "cursor_field": "created", "ingestion_type": "cdc"},
+    "tasks": {"primary_keys": ["id"], "cursor_field": "created", "ingestion_type": "cdc"},
+    # ``users`` has no timestamp field on the actiTIME API (verified
+    # against live response — see class docstring). Treat as a snapshot.
+    "users": {"primary_keys": ["id"], "ingestion_type": "snapshot"},
```
yyoli-db, line 101: There are several things that are inconsistent between the code and README.doc
ingestion_type of the users is one example, and there are several more like the primary keys, schema

please rerun the skills for the doc generation

src/databricks/labs/community_connector/sources/actitime/actitime.py
```diff
@@ -0,0 +1,1058 @@
...
+                return
+            records = self._unwrap_records(body, key="items")
+            if not records:
+                break
+            for raw in records:
+                yield raw if _raw else self._map_record(table_name, raw)
+                emitted += 1
+                if emitted >= max_records:
+                    return
+            if len(records) < page_size:
```
yyoli-db: this can be terminated earlier if the `limit` is configured to be big and the server cpas the number of records per page.

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/actitime/actitime.py
fnapolitano73: I fixed this in bb61503. Regenerated the README from TABLE_METADATA / TABLE_SCHEMAS so the Tables section matches code exactly — users is now snapshot, timetrack PK is (user_id, date, task_id), leavetime PK is (user_id, date, leave_type_id). Also cleaned up the schema  notes (removed non-existent users.user_groups/user_roles rows, fixed userRates.leave_rates from MapType to ArrayType<StructType>).

src/databricks/labs/community_connector/sources/actitime/actitime.py
fnapolitano73: fixed in bb61503. The loop now trusts the envelope's server-applied limit (body["limit"]) instead of the requested page_size, and advances offset by len(records). That way a server cap below the requested limit no longer breaks out after the first page. Filed #178 separately for the adjacent _read_time_window/stopAfter overflow concern since that's a different bug class (no offset/limit on those endpoints — needs window shrinking).

src/databricks/labs/community_connector/sources/actitime/actitime.py
```diff
@@ -0,0 +1,1066 @@
...
+            return
+        record = dict(body)
+        # Synthesise a constant primary key so the row is addressable.
+        record.setdefault("id", 1)
+        yield self._map_record("settings", record)
+
+    def _read_user_rates(self, table_options: dict[str, str]) -> Iterator[dict]:
+        """/userRates is per-user: fan out one request per user from /users."""
+        # Pull a bounded user list. We use offset/limit pagination here too.
+        users = list(self._read_offset_limit("users", table_options, params={}, _raw=True))
```
yyoli-db: This should be minor because we probably won't set batch size to be smaller than number of total user. However, if this could be common, then we need to fix it

`table_options` here is the **userRates** table_configuration. `_read_offset_limit` consults `limit` and `max_records_per_batch` from it. An operator who sets `max_records_per_batch=100` on `userRates` to cap rate rows will inadvertently also cap the `/users` discovery list at 100 users, silently dropping all `userRates` rows for users 101+.

Pass `{}` (or an explicit users-only options dict) to the inner call

src/databricks/labs/community_connector/sources/actitime/actitime.py
```diff
@@ -0,0 +1,1066 @@
...
+        window_days = max(1, int(table_options.get("window_days", "7")))
+        stop_after = int(table_options.get("stopAfter", "1000"))
+
+        window_end_date = min(
+            cursor_date + timedelta(days=window_days), self._init_date
+        )
+        # The actiTIME range is inclusive of both endpoints, so we request
+        # [read_from, window_end_date - 1 day] to avoid double-fetching the
+        # boundary day across consecutive windows.
+        window_end_inclusive = max(window_end_date - timedelta(days=1), read_from)
```
yyoli-db: When the cursor reaches the final window, `window_end_date = min(cursor + window_days, self._init_date) = self._init_date`, then `window_end_inclusive = self._init_date - 1`. The current trigger therefore never requests data dated on `self._init_date` itself — those records surface only on the *next* Trigger.AvailableNow invocation (which gets a new `_init_date`).

For nightly ingestion this means same-day rows are always one trigger late. Either read up-to-and-including `self._init_date` on the last window, or document the lag in the README's "Best Practices" section.

src/databricks/labs/community_connector/sources/actitime/actitime.py
```diff
@@ -0,0 +1,1066 @@
...
+        # timetrack: nested envelope flattened to one row per record. The
+        # inner ``records`` array only carries ``taskId`` and ``time`` —
+        # there is no record-level id/comment/approved/locked on this
+        # tenant's API. ``day_offset`` is the day index inside the
+        # response window (0 = dateFrom).
+        "timetrack": StructType(
+            [
+                StructField("user_id", LongType()),
+                StructField("date", DateType()),
+                StructField("day_offset", LongType()),
```
yyoli-db: **`day_offset` is not a stable record property.**

It's the day index *inside the response window* — so the same `(user_id, date, task_id)` fetched in two overlapping windows (which happens routinely under `lookback_days > 0`) emits two rows with different `day_offset` values. Because `timetrack` is `append` with PK `(user_id, date, task_id)`, downstream dedup keeps one — but the column then flickers between runs in a non-deterministic way that looks like an "update" but isn't.

Safest fix is to drop `day_offset` from the emitted schema. If you want a transport-side ordinal for debugging, keep it but document explicitly that it's not stable across runs.

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/actitime/actitime.py
fnapolitano73: Good catch — and on investigation the leak runs deeper than just `table_options` inheritance. Three cliff edges total:

1. **Operator-set cap** (what you flagged): fixed by passing `{}` at both internal call sites (`_read_user_rates` and `_read_time_window`).
2. **Helper default cap**: `max_records_per_batch` defaulted to `"100000"` inside `_read_offset_limit` itself. So even with `{}`, a tenant with >100k users would still hit silent truncation. Changed the default to `sys.maxsize` — opt-in semantics, not a hidden ceiling.
3. **Cursor model interaction (worst)**: `_read_full_list_cdc` and `_read_snapshot` return `{"done": True}` after one call with no resumption. So the 100k cap on a 250k-row table was silently dropping records 100001-250000 *permanently* until the pipeline state was reset.

Fixed in 6eacad1 with three coordinated edits: gate the cap on `not _raw` in the helper (so internal /users walks drain regardless), change the default to `sys.maxsize`, and pass `{}` at both internal sites for page-size hygiene. Two regression tests added that set tiny `max_records_per_batch` on userRates and timetrack and assert the inner /users walk still drains.

src/databricks/labs/community_connector/sources/actitime/actitime.py
fnapolitano73: Fixed in 80d3203. Conditional on `window_end_date == upper_bound` — at the final window we include the boundary; intermediate windows still apply `-1 day` to avoid double-fetching across consecutive windows in the same trigger.

**Cost of the fix is negligible**: one day of double-fetch per trigger, fully absorbed by the existing `lookback_days=7` overlap (which already re-fetches the previous week) and PK-based upsert. Effectively free.

Scope expansion authorized by the PR owner:

- **Mirrored in `_read_approval_status`** at lines 814-818 — identical pattern; keeps the three time-windowed tables consistent.
- **New `forward_days` option** (default 365) for `/leavetime` and `/approvalStatus`: actiTIME supports planned leave (a user enters vacation months ahead) and future-week approvals on leave-only weeks. The previous `_init_date` clamp made these unreachable. `/timetrack` stays clamped (logged hours don't exist in the future) regardless of the setting. Allowlist updated in `connector_spec.yaml`; documented in README's "Time-window options" section.

Tests added: assert `dateTo == _init_date` on timetrack's final window; assert `dateTo == _init_date + N` for leavetime/approvalStatus with `forward_days=N`.

src/databricks/labs/community_connector/sources/actitime/actitime.py
fnapolitano73: Verified empirically against the connector's own live cassette (`tests/unit/sources/actitime/cassettes/TestActitimeConnector.json`):

- **130 / 130** timetrack+leavetime entries match `dayOffset == (entry.date - request.dateFrom).days`. Zero exceptions across the full corpus.
- **19 distinct dates** already return with **different `dayOffset` values across different requests** to the same tenant in our recorded data. Examples: 2026-04-06 flickers between `dayOffset=1` and `dayOffset=2`; 2026-04-13 between 1 and 2; 2026-04-09 between 4 and 5.

So the field is unambiguously transport state, not a record property. The flicker isn't hypothetical — it's already present in our recorded responses.

Dropped entirely in 403871c (timetrack and leavetime schemas, `_map_record`, `_flatten_time_window_batch`, README schema notes). Regression test added that fetches the same `(user_id, date, task_id)` across two windows with different `lookback_days` (so different `dateFrom`) and asserts emitted rows are byte-identical — locks the determinism property.

No schema break: PR is `[needs-live-testing]` and unmerged; no test assertions referenced the column (grepped to confirm); framework handles community-connector schema evolution.

fnapolitano73:
## Second review pass — all three comments resolved

Pushed three commits on `fna_actitime_new`:

| Commit | Comment | Fix |
|---|---|---|
| 6eacad1 | [#3228778226](https://github.com/databrickslabs/lakeflow-community-connectors/pull/176#discussion_r3228778226) — `max_records_per_batch` leak | Gate cap on `not _raw` in helper; default to `sys.maxsize`; pass `{}` at internal sites. Eliminates silent truncation at any tenant size. |
| 80d3203 | [#3228778237](https://github.com/databrickslabs/lakeflow-community-connectors/pull/176#discussion_r3228778237) — final-window boundary | Conditional `-1 day` only on intermediate windows; mirrored in `_read_approval_status`. New `forward_days` option (default 365) for `/leavetime` and `/approvalStatus`. |
| 403871c | [#3228778244](https://github.com/databrickslabs/lakeflow-community-connectors/pull/176#discussion_r3228778244) — `day_offset` instability | Dropped from both schemas. Empirically verified: 130/130 entries match the formula `(date - dateFrom).days`; 19 dates already flicker in our recorded cassette. |

**Tests** (all simulate-mode):
- 10 → **15 passed**, 2 skipped (unchanged: `partition_requires_lakeflow_connect`, `read_table_deletes`).
- 5 new regression tests added covering each fix.
- Pylint: 9.84/10 (was 9.87/10 — small dip from added locals in the new boundary logic).

**Live re-validation** (required before merge):
- `day_offset` removal drifts the projection of recorded responses.
- `forward_days` extends the API request range for `/leavetime` and `/approvalStatus`.

Will refresh the cassette via `/validate-connector actitime` and post the new validation report.

yyoli-db:
let me know when your LPP gets approved and then we could merge the change.

fnapolitano73:
LPP is approved

yyoli-db (approved):

PR #180, opened 2026-05-12
fix(actitime): fan out /timetrack /leavetime by userIds batch (closes #178)

### LINKED ISSUE #178: actitime: `_read_time_window` silently drops overflow when window record count exceeds `stopAfter`

fnapolitano73:
## Summary

Fixes [#178](https://github.com/databrickslabs/lakeflow-community-connectors/issues/178).

`_read_time_window` (used by `timetrack` and `leavetime`) was issuing a single GET per sliding window with `stopAfter=1000` as a server-side result cap, then advancing the cursor to `window_end_date` unconditionally. Any tenant with more than ~100 active users trips the cap on most days and the overflow is dropped silently.

The fix switches to per-user fan-out:

- Enumerate users with the existing `_read_offset_limit("users", ...)` helper (same shape `_read_user_rates` already uses).
- Slice into batches of `user_batch_size` (new option, default `100`).
- Issue one GET per batch with `userIds=u1,u2,…` — `actitime_api_doc.md:267, 321` confirms this filter is supported on both endpoints.
- Per-call response size is now bounded by the batch, not by `stopAfter`.
- `stopAfter` is demoted to an opt-in passthrough — forwarded per batch only when the caller explicitly sets it.

The flatten step is extracted into `_flatten_time_window_batch` to keep the new request loop under pylint's branch limit. Empty-tenant case short-circuits but still advances the cursor so the framework makes progress.

## Tenant scaling

| Tenant size | Calls/microbatch (default `user_batch_size=100`) |
|---|---|
| 50 users | 1 |
| 1000 users | 10 |
| 10000 users | 100 |

## Stacks on #176

This branch is cut from `fna_actitime_new` (PR #176) because the actitime connector itself only lives on that branch. The diff against `master` will include both #176's connector and this fix on top — please merge #176 first, then rebase this PR.

## Test plan

- [x] `pytest tests/unit/sources/actitime/ -q` → 8 passed, 2 skipped (simulate mode)
- [x] `CONNECTOR_TEST_MODE=live pytest …` → 8 passed, 2 skipped against the live actiTIME dev tenant
- [x] Pylint with CI flags → 10.00/10
- [x] Live validation cassette regenerated 2026-05-12 (12 validated / 4 mismatched / 0 missing — same documented framework-limitation 404s)
- [ ] Workspace smoke: re-trigger `actitime_pipeline` in `e2-demo-field-eng` once merged


fnapolitano73:
Folded into #176 per @yyoli-db's review feedback — commit `3264ec7` (per-user fan-out for #178) is now part of PR #176. Closing this one so review stays in a single place.

PR #181, opened 2026-05-12
Add PUBG community connector

DatabricksLuke:
## Summary

Adds a community connector for the [PUBG Developer API](https://documentation.pubg.com/en/introduction.html), ingesting player, match, and game data into Databricks Delta tables via Lakeflow Connect.

## Supported Tables

| Table | Ingestion Type | Primary Key(s) | Description |
|-------|---------------|----------------|-------------|
| `samples` | append | `id` | Random sample of recent match IDs (refreshes every 24h) |
| `leaderboards` | snapshot | `leaderboardId`, `accountId` | Top-500 ranked players per game mode and season |
| `players` | snapshot | `id` | Player profile and match history |
| `matches` | append | `id` | Match metadata (map, mode, duration, rosters) |
| `season_stats` | snapshot | `accountId`, `seasonId` | Per-season ranked stats per player |
| `lifetime_stats` | snapshot | `accountId` | All-time aggregate stats per player |
| `mastery` | snapshot | `accountId`, `masteryType` | Weapon and survival mastery levels |
| `clans` | snapshot | `id` | Clan profile and member list |
| `telemetry` | append | `matchId`, `eventIndex` | Per-match event log (kills, damage, movement) |

## Auth

Bearer JWT via the [PUBG Developer Portal](https://developer.pubg.com). Required connection parameters: `api_key`, `shard`.

## Tests

- **23 mock tests** in `test_pubg_mocked.py` — run in CI, no credentials required
- **Live integration tests** in `test_pubg_lakeflow_connect.py` — excluded from CI via `test_exclude.txt`
- Pipeline validated end-to-end on e2-dogfood (`luke_smith_catalog1.pubg_steam`)

## Files Changed

| File | Description |
|------|-------------|
| `sources/pubg/pubg.py` | Connector implementation |
| `sources/pubg/connector_spec.yaml` | UC connection spec |
| `sources/pubg/README.md` | User-facing documentation |
| `sources/pubg/__init__.py` | Package export |
| `sources/pubg/pyproject.toml` | Package metadata and dependencies |
| `sources/pubg/pubg.svg` | Connector logo |
| `sources/pubg/pubg_api_doc.md` | API research doc |
| `sources/pubg/_generated_pubg_python_source.py` | Merged Python source |
| `tests/unit/sources/pubg/test_pubg_mocked.py` | Mock-based unit tests |
| `tests/unit/sources/pubg/test_pubg_lakeflow_connect.py` | Live integration tests |
| `.github/workflows/test_exclude.txt` | Exclude live tests from CI |


yyoli-db:
It seems we have not file a LPP?  please check http://go/community-connectors and file the LPP like here: http://go/community-connectors/tracking

yyoli-db:
Als please run `/self-review-connector` @DatabricksLuke

.github/workflows/test_exclude.txt
```diff
@@ -1,20 +1,27 @@
...
 tests/unit/sources/sap_successfactors/test_group_runner.py
-
-# Standalone FHIR utility tests have 2 pre-existing failures that are
-# unrelated to the simulator. Re-enable once those are fixed.
+tests/unit/sources/stripe/test_stripe_lakeflow_connect.py
+tests/unit/sources/surveymonkey/test_surveymonkey_lakeflow_connect.py
+tests/unit/sources/zendesk/test_zendesk_lakeflow_connect.py
+tests/unit/sources/zoho_crm/test_zoho_crm_lakeflow_connect.py
+tests/unit/sources/fhir/test_fhir_lakeflow_connect.py
 tests/unit/sources/fhir/test_fhir_utils.py
```
yyoli-db, line 27: **BLOCKING (B1).** This entire file is a revert of master — see PR #157/#158 history. Master's version is a short opt-out list (3 entries: example/sap_successfactors group runner/fhir utils) because the source-simulator framework lets connector tests run offline in CI. This PR re-introduces the legacy 16-entry list, which silently disables CI for `appsflyer`, `azure_devops`, `github`, `gmail`, `google_analytics_aggregated`, `hubspot`, `microsoft_teams`, `mixpanel`, `osipi`, `qualtrics`, `sap_successfactors`, `stripe`, `surveymonkey`, `zendesk`, `zoho_crm`, `fhir`. Almost certainly a rebase artifact from a stale base.

Fix: restore master's file (`git checkout origin/master -- .github/workflows/test_exclude.txt`) and — only if B2 (simulator spec) is not addressed — append one line:
```
tests/unit/sources/pubg/test_pubg_lakeflow_connect.py
```

tests/unit/sources/pubg/test_pubg_lakeflow_connect.py
```diff
@@ -0,0 +1,557 @@
...
+import json
+import logging
+
+import pytest
+
+from databricks.labs.community_connector.sources.pubg.pubg import (
+    SUPPORTED_TABLES,
+    PubgLakeflowConnect,
+)
+from tests.unit.sources.test_suite import LakeflowConnectTests
```
yyoli-db, line 34: **BLOCKING (B2).** This test class subclasses `LakeflowConnectTests` but does **not** set `simulator_source` or `replay_config`, so it can only run against the live API — and since it's listed in `test_exclude.txt`, it never runs in CI at all.

The project's documented developer flow (CLAUDE.md) is to ship a simulator spec at `src/databricks/labs/community_connector/source_simulator/specs/pubg/{endpoints.yaml,corpus/}` and write the test class as:
```python
class TestPubgConnector(LakeflowConnectTests):
    connector_class = PubgLakeflowConnect
    simulator_source = "pubg"
    replay_config = {"api_key": "simulator-fake-jwt", "shard": "steam"}
```
See `tests/unit/sources/shopify/test_shopify_lakeflow_connect.py` (in this same diff) for the canonical example. After that, you can drop the bespoke `_seed_ids_from_api` + `__SEED_FROM_*__` sentinel logic, drop this file from `test_exclude.txt`, and the full harness (offset contract, termination, PK checks) runs offline in CI.

src/databricks/labs/community_connector/sources/pubg/pubg.py
```diff
@@ -0,0 +1,1316 @@
...
+    def _request_with_retry(self, url: str, params: dict | None = None) -> requests.Response:
+        """GET ``url`` with exponential backoff on rate-limit / server errors.
+
+        Respects the ``X-RateLimit-Reset`` header when present on 429 responses
+        so we wait exactly as long as the server requires, rather than sleeping
+        a fixed amount.
+        """
+        backoff = INITIAL_BACKOFF
+        for attempt in range(MAX_RETRIES):
+            resp = self._session.get(url, params=params)
```
yyoli-db, line 452: **MAJOR (M1).** No `timeout=` on the request. Without one, a dead peer or stuck TLS handshake blocks `read_table` indefinitely and the pipeline trigger hangs. Reference `shopify_utils.py:33` uses `timeout=30`.

```suggestion
            resp = self._session.get(url, params=params, timeout=30)
```

Also applies to `_cdn_get` at line 479 — consider a longer telemetry timeout (e.g. 60s) since CDN payloads can be large.

src/databricks/labs/community_connector/sources/pubg/connector_spec.yaml
```diff
@@ -0,0 +1,48 @@
...
+#   matches:       match_ids
+#   season_stats:  player_ids, season_id
+#   lifetime_stats: player_ids
+#   leaderboards:  season_id, game_modes
+#   mastery:       player_ids, mastery_types
+#   clans:         clan_ids
+#   samples:       (none)
+#   telemetry:     match_ids
+# ============================================================================
+external_options_allowlist: "player_names,player_ids,match_ids,season_id,game_modes,mastery_types,clan_ids"
```
yyoli-db, line 48: **MAJOR (M2).** `created_at_start` is documented as a `samples` option in README.md:263 and consumed by `_read_samples` at pubg.py:1111, but it's missing from this allowlist. UC will reject it at runtime.

Either add it:
```yaml
external_options_allowlist: "player_names,player_ids,match_ids,season_id,game_modes,mastery_types,clan_ids,created_at_start"
```
and update the comment on line 45 (`samples: (none)` → `samples: created_at_start`), or drop `created_at_start` from the README + connector since the cursor mechanism already handles incrementality.

src/databricks/labs/community_connector/sources/pubg/pubg.py
```diff
@@ -0,0 +1,1316 @@
...
+            "matchIds": match_ids,
+        }
+
+        end_offset = {"cursor": created_at} if created_at else {}
+
+        # If offset did not advance, there's no new data.
+        if start_offset and start_offset == end_offset:
+            return iter([]), start_offset
+
+        return iter([record]), end_offset
```
yyoli-db, line 1148: **MAJOR (M3).** When `data` is empty (`{"data": {}}`) or `data.id` is missing, this still emits one record with `id=None`. `samples`' primary key is `id`. Append tables don't enforce non-null PKs in `_validate_record_fields`, so the harness doesn't catch it.

Guard:
```python
if not data.get("id"):
    return iter([]), start_offset or {}
```
And tighten `test_read_samples_empty_response` in `test_pubg_mocked.py` to assert an empty iterator for empty `data`.

src/databricks/labs/community_connector/sources/pubg/pubg.py
```diff
@@ -0,0 +1,1316 @@
...
+            if created_at and (last_created_at is None or created_at > last_created_at):
+                last_created_at = created_at
+
+        if not records:
+            return iter([]), start_offset or {}
+
+        end_offset: dict = {"seen_ids": sorted(seen_ids)}
+        if last_created_at:
+            end_offset["cursor"] = last_created_at
+
```
yyoli-db, line 729: **MAJOR (M4).** `matches` is declared `append` with `cursor_field=createdAt`, but the cursor written here is `last_created_at` from the current batch's match IDs — not monotonically advanced from `start_offset`. If a subsequent trigger supplies a different `match_ids` list, the cursor can regress (the new max can be smaller than the previous one), which breaks the implicit monotonicity downstream merge plans rely on.

Either:
1. Drop the cursor from the offset entirely and rely on `seen_ids` (and update metadata to `cursor_field=None`).
2. Clamp:
   ```python
   prev_cursor = start_offset.get("cursor") if start_offset else None
   if prev_cursor and (last_created_at is None or last_created_at < prev_cursor):
       last_created_at = prev_cursor
   ```

Separately: `telemetry` declares `cursor_field=_D` in metadata but the offset returned at line 1234 is `{"seen_ids": [...]}` only — no `_D` cursor ever appears. Pick one and make metadata + offset agree.

src/databricks/labs/community_connector/sources/pubg/pubg.py
```diff
@@ -0,0 +1,1316 @@
...
+        Offset tracks which match IDs have already been processed.
+        """
+        raw_ids = table_options.get("match_ids", "").strip()
+        if not raw_ids:
+            raise ValueError("telemetry table requires 'match_ids' in table_options")
+
+        all_ids = [mid.strip() for mid in raw_ids.split(",") if mid.strip()]
+        seen_ids: set[str] = set(start_offset.get("seen_ids", [])) if start_offset else set()
+
+        import json as _json  # local import to keep module-level imports minimal
```
yyoli-db, line 1170: **MINOR (m1).** Local imports are usually justified to avoid heavy module-level cost; `json` is a stdlib module that's already implicitly imported by `requests`. Move to the top of the file:
```python
import json
```
and drop the `_json` alias (`payload": _json.dumps(event)` → `json.dumps(event)`).

src/databricks/labs/community_connector/sources/pubg/pubg.py
```diff
@@ -0,0 +1,1316 @@
...
+        if resp.status_code != 200:
+            raise RuntimeError(f"Failed to fetch seasons: {resp.text}")
+        for season in resp.json().get("data", []):
+            if season.get("attributes", {}).get("isCurrentSeason"):
+                return season["id"]
+        # Fallback: return the last season in the list.
+        seasons = resp.json().get("data", [])
+        if not seasons:
+            raise RuntimeError("No seasons returned by the API")
+        return seasons[-1]["id"]
```
yyoli-db, line 493: **MINOR (m2).** `resp.json()` is called twice on the success path (line 486 in the for-loop, then line 490 for the fallback). Cache it:
```python
body = resp.json()
for season in body.get("data", []):
    if season.get("attributes", {}).get("isCurrentSeason"):
        return season["id"]
seasons = body.get("data", [])
if not seasons:
    raise RuntimeError("No seasons returned by the API")
return seasons[-1]["id"]
```

src/databricks/labs/community_connector/sources/pubg/pubg.py
```diff
@@ -0,0 +1,1316 @@
...
+                                "killDeathRatio": p_stats.get("killDeathRatio"),
+                                "kda": p_stats.get("kda"),
+                                "averageRank": p_stats.get("averageRank"),
+                            },
+                        }
+                    )
+
+                # Leaderboards max out at 500 players per page; stop after page 1
+                # (the API only has one page of results).
+                break
```
yyoli-db, line 923: **MINOR (m3).** The `while True:` (line 873) and `page` variable are dead scaffolding — the loop unconditionally breaks here after page 1, and `page` is read but never advanced. Either drop the loop entirely (and the `page` variable + `page[number]` param) or actually paginate if the API supports it. The current code reads as if pagination were possible but isn't.

src/databricks/labs/community_connector/sources/pubg/pubg.py
```diff
@@ -0,0 +1,1316 @@
...
+# ---------------------------------------------------------------------------
+# Static schemas
+# ---------------------------------------------------------------------------
+
+# Reusable sub-types
+_STATS_FIELDS = [
+    StructField("assists", IntegerType()),
+    StructField("boosts", IntegerType()),
+    StructField("dBNOs", IntegerType()),
+    StructField("dailyKills", IntegerType()),
```
yyoli-db, line 55: **MINOR (m6).** Casing inconsistency: `_STATS_FIELDS` uses `dBNOs` (lowercase d) here, but `_PARTICIPANT_STATS_TYPE` at line 103 uses `DBNOs` (uppercase D). PUBG's raw API uses both. Pick one in our schema and normalize at parse time — mixing them in the same connector is confusing for downstream consumers writing queries.

src/databricks/labs/community_connector/sources/pubg/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[project]
+name = "lakeflow-community-connectors-pubg"
+version = "0.1.0"
+description = "PUBG connector for Lakeflow Community Connectors"
+requires-python = ">=3.10"
+dependencies = [
+    "pyspark>=3.5.0,<4.0",
+    "pydantic>=2.0.0,<3.0",
+    "requests>=2.28.0,<3.0",
```
yyoli-db, line 13: **Nit.** `pydantic` is declared as a runtime dependency but the PUBG connector never imports it. (This was likely copied from Shopify's `pyproject.toml` which also doesn't appear to use it — that's a separate cleanup.) Drop:
```toml
    "pydantic>=2.0.0,<3.0",
```

tests/unit/sources/pubg/configs/dev_table_config.json
```diff
@@ -0,0 +1,30 @@
+{
+  "samples": {
+    "created_at_start": ""
+  },
+  "leaderboards": {
+    "game_modes": "squad"
+  },
+  "players": {
+    "player_names": "__SEED_FROM_SAMPLES__"
```
yyoli-db, line 9: **Nit.** The `__SEED_FROM_SAMPLES__` / `__SEED_FROM_PLAYERS__` sentinel strings are a clever workaround for the live-API seeding step, but they're bespoke to this connector and rely on `setup_class` filtering them out (test file lines 205–210). Once B2 (simulator spec) lands, this file should become a normal static JSON like Shopify's `dev_table_config.json`, with real fake IDs that the corpus serves.

PR #190, opened 2026-05-22
Add HL7 v2 community connector (GCP Healthcare API + Unity Catalog Volumes)

Adityasekar:
## Summary

- Adds a new **HL7 v2** community connector (`sources/hl7_v2/`) supporting two source modes:
  - **GCP Healthcare API** — reads HL7 v2 messages from a Google Cloud Healthcare HL7v2 store via a sliding time-window read pattern (Strategy A)
  - **Unity Catalog Volume** — reads HL7 v2 files directly from a UC Volume path that Spark executors FUSE-mount at runtime. No SDK, no HTTP, no auth options beyond the executor's UC identity (`READ VOLUME` grant). Same FUSE pattern the `dicomweb` connector uses on the write side.
- Implements all 23 standard HL7 v2 segment tables: MSH, EVN, PID, PD1, PV1, PV2, NK1, MRG, AL1, IAM, DG1, PR1, ORC, OBR, OBX, NTE, SPM, IN1, GT1, FT1, RXA, SCH, TXA
- Lossless extraction of all composite HL7 data types (CWE, CX, XPN, XCN, XON, XAD, XTN, EI, EIP, TQ, DLN, SPS, MOC) with full sub-component flattening
- Full test suite: **345 passed, 4 skipped** (simulate mode); live cassette validated against real GCP source with 0 drift
- Pylint **10.00/10**; self-review score **100/100 — READY**

## Artifacts

| File | Description |
|---|---|
| `sources/hl7_v2/hl7_v2.py` | Main connector class (GCP + Volume modes) |
| `sources/hl7_v2/hl7_v2_composites.py` | Composite data type extractors (CWE, CX, XPN, …) |
| `sources/hl7_v2/hl7_v2_extractors.py` | Per-segment flatten functions |
| `sources/hl7_v2/hl7_v2_schemas.py` | PySpark StructType schemas for all 23 tables |
| `sources/hl7_v2/connector_spec.yaml` | Connection parameters spec (GCP + Volume auth methods) |
| `sources/hl7_v2/README.md` | End-user quickstart and parameter reference |
| `sources/hl7_v2/hl7_v2_api_doc.md` | Full API doc (all 23 segment tables, all fields) |
| `source_simulator/specs/hl7_v2/` | Simulator spec, handlers (GCP only — Volume mode is HTTP-less), and corpus |
| `tests/unit/sources/hl7_v2/` | Test suite — one test class per source mode (`TestHL7V2Connector` for GCP, `TestHL7V2VolumeConnector` for Volume), plus per-segment test files |

## Test plan

- [x] `pytest tests/unit/sources/hl7_v2/ -q` — 345 passed, 4 skipped
- [x] Pylint 10.00/10 on all connector source files
- [x] Live GCP record run (2026-05-25): 2 endpoints validated, 48 calls, 0 drift
- [x] Volume mode: 7 sample HL7 files staged in a class-scoped temp dir with mtimes matching the GCP corpus createTimes — cursor walks identically in both modes
- [x] Self-review: 100/100 — READY (all 51 checks pass)

## Changes since first review

This PR was originally opened with a Delta-table source mode. Based on review feedback, the Delta path was removed and replaced with a simpler Unity Catalog Volume mode that reads HL7 files directly from FUSE-mounted UC Volumes — no SDK, no Databricks SQL warehouse, no extra HTTP hops.


Adityasekar:
## Self-review update — changes since the last review

In addition to addressing **all** of @yyoli-db's review comments, this revision
aligns the connector's schema design with the convention used across similar
connectors in this repo — most directly **FHIR**, the logical healthcare
equivalent. Rather than flattening each composite HL7 v2 data type (CWE, CX,
XPN, XCN, XON, …) into many individual top-level columns, single-occurrence
composites are now emitted as **nested STRUCT types** under a single column.
This mirrors how composite/nested data types are represented in the other
connectors, keeps each segment table compact and self-describing, and avoids
column-count explosion.

**Summary of what changed in this revision:**
- ✅ Addressed all reviewer feedback (2 MAJOR + 3 MINOR) — see "Reviewer follow-up" below.
- ✅ **Composite data types → nested STRUCT** (was flattened columns), following the FHIR connector's precedent for composite types.
- ✅ Moved `pyspark` from runtime deps to the `dev` extra — matches the root `pyproject.toml` and every other connector (avoids cluster pip-install conflicts).
- ✅ Guarded the 3 simulator-corpus Z-segment tests so they **skip in live mode** rather than asserting corpus-only content (they pass as before in simulate/CI; this is a test-scoping fix, not a behavior change).
- ✅ Refreshed the live GCP cassette + spec validation — clean, 0 drift.
- ✅ Corrected the `birth_order` type label in the API doc (`IntegerType` → `LongType` / `bigint`).

Self-review score: **99 / 100 — READY** (the only remaining −1 is an intentional match to repo convention on the `lakeflow-community-connectors` pin).

---

# Self-Review — hl7_v2

**Overall: 99 / 100 — READY** *(C10 blocker + B11 stale-cassette + E7 doc label all resolved 2026-06-16; remaining −1 is D10, an intentional match to repo convention)*

Run at: 2026-06-16T09:10:00Z · Updated: 2026-06-16T10:32:00Z (post C10 fix)

pylint: **10.00/10** on all 7 connector files · pytest: **357 passed, 4 skipped** (simulate mode) · live: **13 passed, 6 skipped, 0 failed** with clean spec validation. The two things explicitly worried about — pylint and the test suite — are both green. The C10 packaging blocker and the B11 stale-cassette item have both been resolved.

## Top recommendations

1. ✅ **RESOLVED — BLOCKER (C10)** — `pyproject.toml` declared `pyspark` as a **runtime** dependency. Fixed on 2026-06-16: `pyspark>=4.2.0.dev0` moved from `[project.dependencies]` to `[project.optional-dependencies].dev`, matching the root `pyproject.toml` and all 22 other connectors. Runtime deps are now `lakeflow-community-connectors`, `google-auth`, `requests`. Re-verified: TOML parses, tests still 357 passed / 4 skipped.
2. ✅ **RESOLVED — MAJOR (B11)** — Live cassette refreshed against the GCP store on 2026-06-16 (`cassettes/TestHL7V2Connector.json` + `.coverage.json` + `.validation.json`, all mtime `2026-06-16T16:20`). Validation is **clean**: 2/2 endpoints validated, 0 mismatched, 0 missing-spec (50 message calls + 1 token call, 0 issues). Run with `CONNECTOR_TEST_MODE=live` (note: `record` refreshes the cassette+coverage but skips the spec validator — use `live` to also regenerate `validation.json`). Live run is clean: **13 passed, 6 skipped, 0 failed**. As part of this, 3 corpus-dependent Z-segment tests (`test_z_segment_read_extracts_rows_from_corpus`, `_distinct_custom_segment`, `_rows_omit_set_id_and_match_schema`) were gated to simulate-mode only — they asserted simulator-corpus content (`ZPI`/`ZIN`) absent from the live store and were failing every live run.
3. **MINOR (D10)** — Unbounded pin remaining in `pyproject.toml`: `lakeflow-community-connectors>=0.1.0` has no upper bound. `google-auth` and `requests` are correctly bounded; `pyspark` is now dev-only. *(Note: the unbounded `lakeflow-community-connectors` floor matches what every other connector ships, so this is optional polish.)*
4. ✅ **RESOLVED — MINOR (E3/E7)** — `hl7_v2_api_doc.md:228` `birth_order` type label corrected `IntegerType` → `LongType` (matches `_int_field` at `hl7_v2_schemas.py:882`).

**Remaining −1:** only D10 (rec #3) — the unbounded `lakeflow-community-connectors>=0.1.0` pin. Left as-is intentionally: all 22 other connectors ship it unbounded, so bounding it would make `hl7_v2` the lone divergent connector. Not recommended to "fix."

The two reviewer MAJORs and three MINORs from `yyoli-db` (Jun 11) are all addressed in code and covered by regression tests — see "Reviewer follow-up" at the bottom.

## A. Connector implementation — 25 / 25
- ✅ A1. Class extends `LakeflowConnect` — `hl7_v2.py:135`
- ✅ A2. All abstract methods present — `list_tables` `:347`, `get_table_schema` `:350`, `read_table_metadata` `:355`, `read_table` `:370`
- ✅ A3. `read_table_deletes` not required — `ingestion_type: "append"` on every table, no `cdc_with_deletes` (`hl7_v2.py:365`)
- ✅ A4. Read pattern: deliberate sliding time-window (Strategy A) — GCP `_fetch_messages_in_window` `:503`, Volume `_fetch_messages_from_volume` `:600`
- ✅ A5. `_init_ts` cap present — set `:184`, early-return guard `:404-405`, right-side `min(window_end, _init_ts)` `:413-416`
- ✅ A6. `max_records_per_batch` honored `:408`; deliberate client-side cut at message boundary with `create_time` sort + cursor rewind `:442-455`. Server-side limit (Strategy B) not applicable — GCP filter+pagination has no top-N under `createTime` ordering; Volume has no server.
- ✅ A7. Sole HTTP call site has `timeout=_REQUEST_TIMEOUT` — `hl7_v2.py:331` (`gcpreader.py:120` dev tool also passes `timeout=30`)
- ✅ A8. Connector source uses `LongType` (`hl7_v2_schemas.py:46,68`); `IntegerType`/`MapType` appear only in the generated bundle, not the source
- ✅ A9. `get_table_schema` returns `StructType` from `get_schema(...)` `:350-353`; the refactor *nests* composites into `StructType` (opposite of flattening)
- ✅ A10. Imports clean — `google.auth`/`google.oauth2` (pinned), `requests`, `pyspark`, std-lib, and local modules only (`hl7_v2.py:29-83`)
- ✅ A11. **Pylint 10.00/10** across all 7 connector source files with the CI flag set
- ✅ A12. No experimental ingestion-agent surface (`SupportsIngestionAgent` / `AgentOperation` / `agent_operations()` absent)

- ✅ B1. Test files exist — `test_hl7_v2_lakeflow_connect.py` (GCP) + `test_hl7_v2_volume_lakeflow_connect.py` (Volume)
- ✅ B2. Both subclass `LakeflowConnectTests` — `test_hl7_v2_lakeflow_connect.py:58`, `test_hl7_v2_volume_lakeflow_connect.py:54`
- ✅ B3. `connector_class = HL7V2LakeflowConnect` set on both
- ✅ B4. `simulator_source = "hl7_v2"` set on both
- ✅ B5. `_replay_config` keys cover every `options.get(...)` — GCP: `source_type, project_id, location, dataset_id, hl7v2_store_id, service_account_json`; Volume: `source_type, volume_path, file_glob`
- ✅ B6. `endpoints.yaml` exists in spec dir
- ✅ B7. Corpus present — `corpus/messages.json` (8 rows)
- ✅ B8. Every connector URL maps to a spec entry — `/token` (OAuth) + `/v1/projects/.../messages` (GET); Volume mode makes no HTTP calls
- ✅ B9. `test_read_terminates` passes (simulate, both modes)
- ✅ B10. All tests pass — **357 passed, 4 skipped** in ~1.1s
- ✅ B11. Record/live validation **refreshed 2026-06-16T16:20** against the live GCP store — clean: 2/2 endpoints validated, 0 mismatched, 0 missing-spec, 0 issues (51 calls). Reflects the current connector (post struct-refactor + review fixes).
- ✅ B12. Coverage 100% — both `/token` and `/v1/.../messages` exercised (`*.coverage.json`), freshly regenerated
- ✅ B13. `synthesize_future_records: {cursor_field: createTime, count: 3}` present — `endpoints.yaml:34-36`
- ✅ B14. `test_utils_class` not required — connector is read-only (no `list_insertable_tables`)

## C. Artifacts — 28 / 28
- ✅ C1. `hl7_v2.py` exists; imports + `py_compile` clean (test suite imports it)
- ✅ C2. `hl7_v2_api_doc.md` exists (1365 lines); covers all 23 segment tables (MSH…TXA headers)
- ✅ C3. `connector_spec.yaml` parses; declares `gcp` + `volume` auth methods, shared `source_type`, and `external_options_allowlist`
- ✅ C4. `README.md` exists (585 lines); covers both source modes, all parameters, and all tables
- ✅ C5. `pyproject.toml` parses; `dependencies` include `requests` and `google-auth` (matches imports)
- ✅ C6. `_generated_hl7_v2_python_source.py` mtime (`2026-06-16T14:36`) ≥ source mtimes — in sync
- ✅ C7. `endpoints.yaml` parses / loadable
- ✅ C8. `corpus/messages.json` is valid JSON (8 rows)
- ✅ C9. Test classes exist (covered by B1)
- ✅ C10. `pyspark` is **not** in runtime `[project.dependencies]` — moved to `[project.optional-dependencies].dev` on 2026-06-16 (fix verified). Runtime deps: `lakeflow-community-connectors`, `google-auth`, `requests`.

## D. Security — 21 / 22
- ✅ D1. No hardcoded secrets (`service_account_json` is a connection option, not a literal)
- ✅ D2. No `eval` / `exec` / dynamic `compile` (the lone match is `re.compile`, a regex)
- ✅ D3. No `subprocess` / `os.system` / `os.popen` / `shell=True`
- ✅ D4. No `verify=False`
- ✅ D5. No `pickle.load(s)`
- ✅ D6. No `yaml.load(` (connector parses no YAML)
- ✅ D7. `volume_path` validated as absolute before use — `hl7_v2.py:246-250`; UC enforces actual `READ VOLUME` at runtime
- ✅ D8. No secrets in logs — `_LOGGER` calls carry paths/offsets, never tokens/headers
- ✅ D9. No `http://` in production paths (HTTPS only)
- ⚠️ D10. One unbounded pin remaining: `lakeflow-community-connectors>=0.1.0`. `google-auth`/`requests` bounded; `pyspark` now dev-only. Matches every other connector, so optional. — *see Top recommendation #3*

- ✅ E1. `list_tables()` (23) ≡ API-doc tables ≡ README tables — `msh, evn, pid, pd1, pv1, pv2, nk1, mrg, al1, iam, dg1, pr1, orc, obr, obx, nte, spm, in1, gt1, ft1, rxa, sch, txa`
- ✅ E2. Connection-parameter keys match across code (`__init__`) / `connector_spec.yaml` / README — both GCP and Volume modes
- ✅ E3. `get_table_schema` column **names** are a subset of the API doc; the one inaccurate *type* label (`birth_order`) has been corrected (see E7)
- ✅ E4. Primary keys consistent — single-segment tables → `[message_id]`; known multi-segment → `[message_id, set_id]`; unknown/Z → `[message_id]`
- ✅ E5. `external_options_allowlist` = `segment_type,window_seconds,start_timestamp,max_records_per_batch` (`connector_spec.yaml:126`) ≡ keys read via `table_options.get(...)` (`hl7_v2.py:352,388,407,408`)
- ✅ E6. README quick-start uses parameter keys that match `connector_spec.yaml`
- ✅ E7. `hl7_v2_api_doc.md:228` `birth_order` corrected to `LongType`. README line 3's "its own Delta table" phrasing is accurate as a *destination* (all Lakeflow connectors land in Delta), not the removed Delta *source* mode, so left as-is.

## Reviewer follow-up (yyoli-db, Jun 11) — all addressed
- **MAJOR 1** (GCP capped-batch rewind drop/dupe under `sendTime` order): fixed by sorting parsed records by `(create_time, source_file)` before the cut — `hl7_v2.py:433`. Regression tests: `test_capped_batch_sorts_by_create_time`, `test_capped_batch_no_drops_or_dupes_across_batches`, `test_capped_batch_keeps_equal_create_time_group_together`.
- **MAJOR 2** (Z-segment `set_id` not in `GENERIC_SEGMENT_SCHEMA`): fixed — `set_id` injected only for known multi-segment tables (`is_known and not single`) — `hl7_v2.py:700-707`. Regression test: `test_z_segment_rows_omit_set_id_and_match_schema`.
- **MINOR** (lexicographic cursor compare on unvalidated `start_timestamp`): fixed via `_normalize_timestamp` — `hl7_v2.py:530-543`. Tests: `test_normalize_timestamp_canonicalizes_user_input`, `test_start_timestamp_option_is_normalized_before_use`.
- **MINOR** (README "send time" vs `create_time`): fixed — README:3 now says `createTime` (GCP) / file mtime (Volume).
- **MINOR** (`GENERIC_SEGMENT_SCHEMA` / `set_id`): resolved with MAJOR 2 — schema kept as-is, `set_id` injection skipped for unknown segments.

yyoli-db:
@Adityasekar 
Could you explain more on why do we want a mode to ingest from Databricks Delta table? I don't feel it the right approach to ingest data from a delta table via Python Data Source into databricks. There are other simpler and more efficient approach

Adityasekar:
@yyoli-db - Fair point, and the pivot makes sense. Quick context: HL7 v2 is fundamentally a messaging protocol, not an API. Most healthcare systems don't expose a query surface like GCP's Healthcare API — they emit MLLP / TCP streams or drop files into a landing zone. The original Delta mode was an attempt to handle the "messages already landed somewhere by an upstream gateway" pattern, but you're right that going through a SQL warehouse for it is overkill.

I've replaced it with a Unity Catalog Volume source mode that reads HL7 files straight off a FUSE-mounted UC Volume — no SDK, no PAT, no warehouse, no extra HTTP. Permissions follow READ VOLUME grants. This mirrors the pattern the dicomweb connector uses on the write side (DICOM files written to a Volume with WRITE FILES) — this is the read-side equivalent. See the updated PR description ("Changes since first review") for the swap.

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2.py
```diff
@@ -0,0 +1,675 @@
...
+        # filter is strict ``createTime > since``.
+        if max_records is not None and len(records) > max_records:
+            cut = max_records
+            # Walk forward until create_time changes — keep all rows
+            # belonging to the message that straddles ``max_records``.
+            straddle_ct = records[cut - 1].get("create_time")
+            while cut < len(records) and records[cut].get("create_time") == straddle_ct:
+                cut += 1
+            records = records[:cut]
+            last_ct = records[-1].get("create_time") if records else None
```
yyoli-db, line 450: **MAJOR (data correctness): GCP `max_records_per_batch` rewind can drop or duplicate records.**

The cursor field is `create_time` and the window filter is on `createTime` (line 492), but GCP fetches are ordered by `sendTime asc` (line 501) and the GCP records are never re-sorted by `create_time`. (The volume path *does* sort by `(createTime, name)` at line 609 — GCP does not.)

This rewind logic assumes `create_time` ordering:
- The straddle loop (lines 434-436) only coalesces **contiguous** equal-`create_time` rows. Under `sendTime` ordering, messages sharing a `create_time` may be non-contiguous, so the cut can split a same-`create_time` group; the next batch's `createTime > last_ct` filter then **silently drops** the un-emitted members → data loss.
- `last_ct = records[-1].create_time` is not the max `create_time` emitted under `sendTime` order, so already-emitted messages with `create_time > last_ct` get re-fetched → **duplication**.

Fix: sort the GCP records by `create_time` (+ stable tiebreaker) before the cut — matching line 609 — or change the GCP `orderBy` to `createTime asc`. The non-capped path is unaffected.

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2.py
```diff
@@ -0,0 +1,675 @@
...
+
+                if segment_type == "MSH":
+                    if msh is not None:
+                        records.append(meta | _extract_msh(msh) | {"raw_segment": msh.raw_line})
+                else:
+                    extractor = _EXTRACTORS.get(segment_type.lower(), _extract_generic)
+                    is_multi_segment = segment_type.lower() not in _SINGLE_SEGMENT_TABLES
+                    for idx, seg in enumerate(msg.get_segments(segment_type), start=1):
+                        row = meta | extractor(seg) | {"raw_segment": seg.raw_line}
+                        if is_multi_segment and "set_id" not in row:
```
yyoli-db, line 706: **MAJOR: Z-segment rows inject `set_id` not present in `GENERIC_SEGMENT_SCHEMA`.**

A custom/Z-segment table is reachable via `table_options['segment_type']` (see `_validate_table`, lines 454-460), and `get_schema` returns `GENERIC_SEGMENT_SCHEMA` for it. That schema declares only metadata + `segment_type` + `field_1..field_25` — no `set_id`. But Z-segments are not in `_SINGLE_SEGMENT_TABLES`, so `is_multi_segment` is True here and `row['set_id'] = idx` is injected. The emitted row then carries a key absent from the declared schema (dropped or errors depending on the framework's row→Arrow strictness). It's also inconsistent with `read_table_metadata`, which uses `primary_keys=['message_id']` (no `set_id`) for the unknown-segment case.

The existing Z-segment tests inspect raw `read_table` dicts and never exercise schema coercion, and the base suite doesn't iterate Z-tables, so this is untested. Fix: skip the `set_id` injection when the segment is unknown/generic (preferred, matches the PK choice), or add `set_id` to `GENERIC_SEGMENT_SCHEMA`.

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2.py
```diff
@@ -0,0 +1,675 @@
...
+            since = self._peek_oldest_create_time()
+        if not since:
+            _LOGGER.info(
+                "read_table(%s): no start cursor resolved "
+                "(start_offset=%s, source_type=%s); returning empty.",
+                table_name, start_offset, self._source_type,
+            )
+            return iter([]), start_offset or {}
+
+        if since >= self._init_ts:
```
yyoli-db, line 404: **MINOR: cursor comparison relies on lexicographic timestamp ordering.**

`since >= self._init_ts` is a string compare; it is only correct when both sides use the exact `%Y-%m-%dT%H:%M:%SZ` shape. Internally-generated cursors do, but the user-supplied `start_timestamp` table option flows in unvalidated (line 388) and is later parsed with `datetime.fromisoformat(since.replace('Z','+00:00'))` (line 406), which accepts shapes (`2024-01-01`, `+00:00` offsets, fractional seconds) that would compare incorrectly here. Consider normalizing `start_timestamp` through `fromisoformat` + reformat before using it.

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2_schemas.py
```diff
@@ -0,0 +1,2240 @@
...
+)
+
+# ---------------------------------------------------------------------------
+# Generic segment schema — used for Z-segments and any unknown segments.
+# Returns segment_type + up to 25 field_N columns.
+# ---------------------------------------------------------------------------
+
+_GENERIC_FIELD_COUNT = 25
+
+GENERIC_SEGMENT_SCHEMA = StructType(
```
yyoli-db, line 2058: Related to MAJOR 2 on `hl7_v2.py`: `GENERIC_SEGMENT_SCHEMA` has no `set_id` column, yet `read_table` injects `set_id` for Z-segment rows (which are multi-segment). Either add `set_id` here or skip the injection for unknown segments so the row shape matches this schema.

src/databricks/labs/community_connector/sources/hl7_v2/README.md
```diff
@@ -0,0 +1,579 @@
+# Lakeflow HL7 v2 Community Connector
+
+The Lakeflow HL7 v2 Connector parses HL7 v2 pipe-delimited messages and loads each segment type into its own Delta table, enabling structured analytics over clinical data. The connector supports incremental append-only ingestion, automatically tracking new messages by their send time.
```
yyoli-db: **MINOR (doc accuracy):** this says messages are tracked by **send time**, but the cursor field is `create_time` (`read_table_metadata` returns `cursor_field: 'create_time'`, and the window filter is on `createTime`). `sendTime` is only the GCP `orderBy`. Suggest: tracked by `createTime` (GCP) / file mtime (Volume).

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/hl7_v2/README.md
Adityasekar: Fixed

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2_schemas.py
Adityasekar: Resolved with comment 2 — kept the schema as-is and stopped injecting set_id for unknown segments.

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2.py
Adityasekar: set_id is now only added for known multi-segment tables (is_known AND not single-segment). Z-segments never get it — even when they repeat — since is_known fails first, so their rows match GENERIC_SEGMENT_SCHEMA and the message_id-only PK. Test: test_z_segment_rows_omit_set_id_and_match_schema.

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2.py
Adityasekar: New _normalize_timestamp canonicalizes start_timestamp (bare date / offset / fractional seconds → %Y-%m-%dT%H:%M:%SZ) before it's compared or used in the filter.

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2.py
Adityasekar: Fixed

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2_extractors.py
```diff
@@ -0,0 +1,1074 @@
...
+        **_cwe_fields(seg, 51, "visit_indicator", repeating=False),
+        **_xcn_array_fields(seg, 52, "other_healthcare_provider"),
+        "service_episode_description": _v(seg.get_field(53)),
+        **_cx_fields(seg, 54, "service_episode_identifier", repeating=False),
+    }
+
+
+def _extract_obr(seg: HL7Segment) -> dict:
+    return {
+        "set_id": _i(seg.get_field(1)) or 1,
```
yyoli-db: Non-blocking (MINOR 3): `_i(seg.get_field(1)) or 1` rewrites a literal `set_id` of `"0"` to `1`, because `_i("0")` is `0` (falsy). HL7 set-ids are 1-based so this is harmless in practice; flagging for awareness — could use an explicit `None`-check if you want to preserve a genuine 0. Same idiom recurs at a few other extractors.

tests/unit/sources/hl7_v2/gcpreader.py
yyoli-db, line 1: please move this file to test dir and in particular, do not include it as part of the package.

yyoli-db (commented):
LGTM but please address the comment of not including the script in the source package. 
and also please run your connector e2e in a real databricks workspace

src/databricks/labs/community_connector/sources/hl7_v2/hl7_v2_extractors.py
Adityasekar: Resolved

tests/unit/sources/hl7_v2/gcpreader.py
Adityasekar: Done

yyoli-db (approved):

PR #191, opened 2026-05-22
Switch community-connector CLI to COMMUNITY connection type with auth_type modes

yyoli-db:
## Summary

- Swap `GENERIC_LAKEFLOW_CONNECT` → `COMMUNITY` across the community-connector CLI and example READMEs.
- Add `--auth-type {static|m2m|u2m|u2m_per_user}` to `create_connection`. Non-static modes validate the fixed required option set per mode and stamp `community_oauth_flow` automatically. `u2m` runs an RFC 6749 / RFC 8252 authorization-code + PKCE flow against a `127.0.0.1` loopback redirect (`--redirect-port` to pin), opens the browser to the consent screen, captures the code, and injects `authorization_code` / `pkce_verifier` / `oauth_redirect_uri`.
- Extend `connector_spec.yaml` with an optional `connection.oauth` defaults block. For non-static auth modes the CLI auto-populates standard fields (`authorization_endpoint`, `token_endpoint`, `oauth_scope`, ...) from the spec so users only have to supply per-app values (`client_id`, `client_secret`). User-supplied `--options` still win.
- Force `profile=DEFAULT` when `DATABRICKS_CONFIG_PROFILE` is unset so the SDK does not silently fall over when `~/.databrickscfg` contains multiple profiles pointing at the same workspace host.

## Example: Gmail end-user invocation

```bash
DATABRICKS_CONFIG_PROFILE=<your-profile> \
community-connector create_connection gmail gmail_connector \
  --auth-type u2m \
  --redirect-port 6969 \
  -o '{"client_id":"<id>.apps.googleusercontent.com","client_secret":"<secret>"}'
```

`authorization_endpoint`, `token_endpoint`, and the recommended `oauth_scope` are filled in from `sources/gmail/connector_spec.yaml`.

## Notes

- The gmail `connector_spec.yaml` change overlaps with #189 (OAuth-aware Gmail connector). When one merges, the other will need a quick rebase on that file.
- Known follow-up: pass `access_type=offline` / `prompt=consent` on the Google authorization URL so UC reliably gets a refresh token (otherwise the connection breaks after the 1h access-token lifetime). Plumbing for spec-declared `authorization_params` is the natural next change.

## Test plan

- [x] Full pytest suite passes (186 tests, including 11 new ones)
- [x] Manual: create a real Gmail UC connection on a workspace using `--auth-type u2m`
- [x] Manual: create an `--auth-type static` connection against an existing static-credential connector
- [x] Manual: confirm `DATABRICKS_CONFIG_PROFILE` override + DEFAULT fallback behavior against a multi-profile `~/.databrickscfg`


tools/community_connector/src/databricks/labs/community_connector_cli/cli.py
```diff
@@ -43,6 +55,23 @@
...
+
+    The SDK already honors ``DATABRICKS_CONFIG_PROFILE`` directly. The only
+    gap this helper fills is the case where that env var is *not* set and
+    ``~/.databrickscfg`` holds multiple profiles pointing at the same
+    workspace host — the SDK cannot auto-pick one, so we explicitly select
+    ``DEFAULT`` to keep the CLI deterministic.
+    """
+    if os.environ.get("DATABRICKS_CONFIG_PROFILE"):
+        return WorkspaceClient()
+    return WorkspaceClient(profile="DEFAULT")
```
yyoli-db: **[P1] DEFAULT-profile fallback regresses env-var auth users whose `~/.databrickscfg` lacks a `[DEFAULT]` section.**

The SDK's `_known_file_config_loader` (databricks/sdk/config.py) only skips config-file loading when `self.profile` is unset. When we pass an explicit `profile="DEFAULT"`, the SDK falls into the file-loading branch unconditionally. If `~/.databrickscfg` exists but only has named profiles (no `[DEFAULT]` section), it raises:

> `ValueError: resolve: ~/.databrickscfg has no DEFAULT profile configured`

Affected users: anyone authenticating purely via `DATABRICKS_HOST`+`DATABRICKS_TOKEN` env vars who also has a `~/.databrickscfg` (from other tooling) with only non-DEFAULT profiles. Today (pre-PR) the SDK takes the env-var path; post-PR they hit a hard error.

The existing test `test_falls_back_to_default_profile` doesn't catch this because it mocks `WorkspaceClient` — it only asserts the kwarg we pass, not the SDK's downstream behavior.

Suggested minimal fix:

```python
if os.environ.get("DATABRICKS_CONFIG_PROFILE") or os.environ.get("DATABRICKS_HOST"):
    return WorkspaceClient()
return WorkspaceClient(profile="DEFAULT")
```

This keeps the new "multiple profiles, same host → force DEFAULT" benefit while preserving env-var-only auth.

tools/community_connector/src/databricks/labs/community_connector_cli/oauth_flow.py
```diff
@@ -0,0 +1,247 @@
...
+        "authorization_endpoint",
+        "token_endpoint",
+    ),
+    AUTH_TYPE_U2M_PER_USER: (
+        "client_id",
+        "client_secret",
+        "authorization_endpoint",
+        "token_endpoint",
+    ),
+}
```
yyoli-db: **[Test gap]** `AUTH_TYPE_U2M` and `AUTH_TYPE_U2M_PER_USER` share the same required-options tuple and the same `_apply_auth_type` missing-fields check (cli.py:347-353). The m2m branch has explicit coverage via `test_create_connection_m2m_requires_oauth_fields`; the u2m branch goes through the identical code but isn't directly tested. Low risk because of the shared code path, but a one-line parametrized test would close the symmetry.

PR #214, opened 2026-06-04
Support cluster_by in table_configuration

yyoli-db:
## Summary
- Add `cluster_by` as a destination-side special key in `table_configuration`. The pipeline reads it from the spec and passes it through to `sdp.create_streaming_table(name=..., cluster_by=[...])` for CDC, snapshot, and append flows.
- Strip `cluster_by` from `get_table_configuration()` so it is never forwarded to the source's `.options(...)` — same pattern as the existing `scd_type` / `primary_keys` / `sequence_by` keys. As a result, **no per-connector `external_options_allowlist` change is needed**.
- Accept `cluster_by` as a list, a JSON-array string, a comma-separated string, or a single column string.

Example usage in a pipeline spec:

```yaml
objects:
  - table:
      source_table: events
      table_configuration:
        cluster_by: ["customer_id", "event_date"]
```

This produces, at pipeline runtime:

```python
sdp.create_streaming_table(name="events", cluster_by=["customer_id", "event_date"])
```

## Test plan
- [x] `tests/unit/libs/test_spec_parser.py` — `cluster_by` is stripped from `get_table_configuration`; `get_cluster_by` parses list / single string / comma string / JSON-array string; returns `None` when absent.
- [x] `tests/unit/pipeline/test_ingestion_pipeline.py` — `cluster_by` is forwarded to `sdp.create_streaming_table` for CDC, snapshot, and append flows; omitted when not specified; stripped from the source's `.options(...)` call.
- [x] Full suite: `tests/unit/libs` + `tests/unit/pipeline` — 324 passed.

### DIFF
--- a/src/databricks/labs/community_connector/libs/spec_parser.py
+++ b/src/databricks/labs/community_connector/libs/spec_parser.py
@@ -18,6 +18,7 @@ from pydantic import (
 SCD_TYPE = "scd_type"
 PRIMARY_KEYS = "primary_keys"
 SEQUENCE_BY = "sequence_by"
+CLUSTER_BY = "cluster_by"
 
 # Valid SCD type values
 SCD_TYPE_1 = "SCD_TYPE_1"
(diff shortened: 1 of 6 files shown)


sheltauro (approved):

erinaldidb (approved):
Awesome work! This is super helpful.

image

PR #218, opened 2026-06-10
Extend connector_spec OAuth shape with regions, flow type, and PKCE; apply to Zoho CRM

rahulkhanna-dbl:
## Summary

Extends the `connector_spec.yaml` OAuth contract so multi-region providers
(Zoho, future SaaS connectors hosted on multiple data centers) and PKCE
requirements can be expressed in the spec rather than the consuming code.
Applies the new shape to Zoho CRM as the first user.

Spec additions (documented in `templates/connector_spec_template.yaml`):

- Connection-param attributes:
  - `default:` — pre-fill value, used when the user submits empty input.
  - `enum:` — restricts the param to a fixed set; renders as a Select in
    the browser form / numbered list in the CLI. Required when the param
    is referenced from `oauth.endpoints_by_region.parameter`.
- `oauth.flow` — `u2m` (default) | `m2m` | `u2m_per_user`. Maps to UC's
  `community_oauth_flow` option.
- `oauth.pkce` — bool, default false. The labs CLI does not yet implement
  PKCE; `pkce: true` raises at resolution time.
- `oauth.endpoints_by_region` — multi-region endpoint map. The `parameter`
  field names a connection param (must be `enum`-typed); `values` keys
  match that param's `enum` exactly. Mutually exclusive with flat
  `authorization_url` + `token_url`.

`tools/scripts/authenticate.py`:

- New `resolve_oauth_endpoints(oauth_cfg, collected)` returns an effective
  oauth config with flat URLs regardless of which form the spec uses; also
  enforces the PKCE-not-supported guard.
- `prompt_for_parameter` honors `enum` (accepts index or value, retries on
  invalid input) and `default` (used on empty input).
- Browser form renders `enum` params as `<select>` and `default` as the
  pre-filled input value.
- Removed the CLI URL/scope override prompts and the corresponding browser
  form fields — the spec is now expressive enough to describe every endpoint
  variant.

Zoho CRM:

- Dropped the `refresh_token` connection param (obtained by the OAuth flow,
  not user-supplied) and the free-form `base_url` (replaced by an enum
  `region` param plumbed through `oauth.endpoints_by_region`).
- `zoho_crm.py` resolves the accounts URL from `region` via a static map;
  `base_url` is kept as a fallback so existing local dev configs keep
  working until users re-run `authenticate.py`.

## Test plan

- [x] `python3.10 -c "import yaml; yaml.safe_load(open('connector_spec.yaml'))"` parses cleanly for Zoho and the template.
- [x] `python3.10 -c "import ast; ast.parse(open('authenticate.py').read())"` syntax-checks.
- [x] `resolve_oauth_endpoints` smoke tests: missing selector → clear error; valid region → expected URLs; unknown region → error; `pkce: true` → error; flat shape resolves; multi-region resolves.
- [x] `prompt_for_parameter` interactive: empty → default; numeric index → value; literal value → value; invalid → retry.
- [x] Browser form: region renders as `<select>` with US `selected`; password fields unchanged.
- [x] `_resolve_accounts_url` in `zoho_crm.py`: case-insensitive region; legacy `base_url` fallback; region wins over base_url; unknown region raises.
- [ ] Re-run `authenticate.py` against the new Zoho spec to regenerate `dev_config.json` with `region` (manual, requires real Zoho OAuth app).

### DIFF
--- a/src/databricks/labs/community_connector/sources/zoho_crm/README.md
+++ b/src/databricks/labs/community_connector/sources/zoho_crm/README.md
@@ -19,8 +19,8 @@ To configure the connector, provide the following parameters in your connector o
 |-----------|------|----------|-------------|---------|
 | `client_id` | string | Yes | OAuth Client ID from Zoho API Console | `1000.XXXXX...` |
 | `client_secret` | string | Yes | OAuth Client Secret from Zoho API Console | `abc123...` |
-| `refresh_token` | string | Yes | OAuth Refresh Token (never expires) | `1000.YYYYY...` |
-| `base_url` | string | No | Zoho accounts URL for your data center | `https://accounts.zoho.com` |
+| `refresh_token` | string | Yes | OAuth Refresh Token (never expires). Obtained automatically via the OAuth flow during connection setup. | `1000.YYYYY...` |
+| `region` | string | Yes | Zoho data center for your account. One of `US`, `EU`, `IN`, `AU`, `CN`, `JP`. Defaults to `US`. | `EU` |
 | `initial_load_start_date` | string | No | Starting point for the first sync. If omitted, syncs all historical data. (ISO 8601 format) | `2024-01-01T00:00:00Z` |
 
 ### Table-Level Options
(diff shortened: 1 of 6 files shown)


src/databricks/labs/community_connector/sources/zoho_crm/connector_spec.yaml
```diff
@@ -56,19 +52,49 @@ connection:
...
-    token_url: "https://accounts.zoho.com/oauth/v2/token"
+    flow: u2m
+    pkce: false
     scopes: "ZohoCRM.modules.ALL,ZohoCRM.settings.ALL"
     extra_auth_params:
-      access_type: "offline"
-      prompt: "consent"
+      access_type: offline
+      prompt: consent
+    endpoints_by_region:
```
yyoli-db, line 81: is this common in oauth? 
what if there is no region? will the `authorization_url`  and `token_url` under `oauth` work?

src/databricks/labs/community_connector/sources/zoho_crm/README.md
```diff
@@ -19,8 +19,8 @@ To configure the connector, provide the following parameters in your connector o
 |-----------|------|----------|-------------|---------|
 | `client_id` | string | Yes | OAuth Client ID from Zoho API Console | `1000.XXXXX...` |
 | `client_secret` | string | Yes | OAuth Client Secret from Zoho API Console | `abc123...` |
-| `refresh_token` | string | Yes | OAuth Refresh Token (never expires) | `1000.YYYYY...` |
-| `base_url` | string | No | Zoho accounts URL for your data center | `https://accounts.zoho.com` |
+| `refresh_token` | string | Yes | OAuth Refresh Token (never expires). Obtained automatically via the OAuth flow during connection setup. | `1000.YYYYY...` |
```
yyoli-db: I think this should removed?

src/databricks/labs/community_connector/sources/zoho_crm/README.md
```diff
@@ -362,7 +362,7 @@ The connector implements automatic retry with exponential backoff for rate limit
 If you see "Token refresh response missing 'access_token'":
 1. Verify your `client_id` and `client_secret` are correct
 2. Check that your `refresh_token` is valid (hasn't been revoked)
```
yyoli-db: this should be removed.

src/databricks/labs/community_connector/sources/zoho_crm/zoho_crm.py
yyoli-db: I believe we no longer pass the refresh_token in because the option of the uc connection won't include the refresh_token anymore.
The `access_token` will be available and should be passed int.

yyoli-db (changes requested):

PR #219, opened 2026-06-15
Register sources via per-source DataSource subclasses + find_data_source()

yyoli-db:
## Summary

Reworks source registration to use PySpark's native `spark.dataSource.register`, replacing the confusing `register(spark, Union[str, DataSource, LakeflowConnect])` overload (which re-imported the connector by FQN inside a dynamically-built wrapper).

Three registration paths:

1. **Direct** (wheel-installed connectors):
   ```python
   from databricks.labs.community_connector.sources.gmail import GmailDataSource
   spark.dataSource.register(GmailDataSource)
   ```
2. **Lookup by name** — new `find_data_source("gmail")` returns the `<Source>DataSource` class.
3. **Legacy** `register(spark, "gmail")` — resolves the merged `_generated_<source>_python_source.py` register function and registers the `lakeflow_connect` format. **No fallback**: it does *not* fall back to `find_data_source`. The intent is to deprecate this path and have users explicitly call `spark.dataSource.register` (path 1/2).

## Changes

- **`LakeflowSource`** gains two class attributes:
  - `_lakeflow_connect_cls` — bound by each per-source subclass. Left `None` on the base so the merged single-file (SDP) path falls back to the `LakeflowConnectImpl` merge placeholder, exactly as today.
  - `_format_name` — overridable, defaults to `"lakeflow_connect"`. `name()` now returns it.
- **`registry.py`** — adds `find_data_source()`; `register()` takes `str | DataSource` only. The string path is the legacy merged-module path **with no `find_data_source` fallback**. The `LakeflowConnect`/FQN-wrapper overload is removed (no callers passed a class — all use the string form).
- **Every source `__init__.py`** (including the new **palantir** connector) exposes a `<Source>DataSource(LakeflowSource)` subclass setting `_lakeflow_connect_cls`. The source-name format override (`# _format_name = "<source>"`) is **included but commented out**, so all sources keep registering as `"lakeflow_connect"`.
- Regenerated all `_generated_*.py` files for the `LakeflowSource` change (CI enforces freshness).
- **Skills**: `implement-connector` now requires exposing the `<Source>DataSource` class (concise instruction pointing at `sources/example`); `self-review-connector` adds **BLOCKER check A13** enforcing that every connector exposes it.

## Format-name decision

- Merged `_generated` path stays `lakeflow_connect` (UC connection-option injection keys off that exact string).
- New subclasses default to `lakeflow_connect` too; the source-name override is left commented for a future change once a source no longer relies on UC injection.

## Notes

- PR #189 (`gmail-improvements`) contains an overlapping version of this refactor bundled with the agent-ops feature; this PR is the standalone foundation off `master` and that PR can rebase on top.

## Test plan

- [x] `tests/unit/sparkpds_registry_test.py` rewritten for the new API (incl. no-fallback assertions) — 8 passed
- [x] `find_data_source` resolves all 22 sources (incl. palantir) to a `LakeflowSource` subclass with `name() == "lakeflow_connect"`
- [x] All source unit suites pass (palantir: 68 passed); `sparkpds_ingestion_agent_test.py` and `test_sparkpds_virtual_tables.py` pass
- [x] `merge_python_source.py all` regenerates deterministically; generated `example`/`palantir` fall back to the substituted `LakeflowConnectImpl` and report `lakeflow_connect`
- [x] pylint 10.00/10 on changed non-generated files

### DIFF
--- a/src/databricks/labs/community_connector/sparkpds/registry.py
+++ b/src/databricks/labs/community_connector/sparkpds/registry.py
@@ -1,171 +1,144 @@
 """
-Registry module for registering LakeflowSource with Spark's DataSource API.
+Registry helpers for Lakeflow source DataSources.
 
-This module provides functions to register a LakeflowConnect implementation
-or a custom DataSource with Spark, making it available as a Python Data Source.
+Three ways to register a source with Spark, all valid:
+
+1. Direct import (preferred for wheel-installed connectors)::
+
+       from databricks.labs.community_connector.sources.gmail import GmailDataSource
+       spark.dataSource.register(GmailDataSource)
+       df = spark.read.format("lakeflow_connect").option(...).load()
+
+2. Lookup by source name (handy when the source is selected at runtime)::
+
+       from databricks.labs.community_connector.sparkpds import find_data_source
+       spark.dataSource.register(find_data_source("gmail"))
+
+3. Legacy ``register()`` (used by pipeline templates)::
+
+       from databricks.labs.community_connector import register
+       register(spark, "gmail")
+
+   This resolves the pre-merged single-file deployment
+   (``_generated_<source>_python_source.py``) and registers the
+   ``lakeflow_connect`` format. It is intentionally limited to that path and
+   will be deprecated — new code should call ``spark.dataSource.register``
+   with the source's ``DataSource`` class (path 1) or :func:`find_data_source`
+   (path 2) instead.
 """
 
 import importlib
 import inspect
 from types import ModuleType
-from typing import Type, Union
-from pyspark.sql import SparkSession
-from pyspark.sql.datasource import DataSource
+from typing import Type
 
-from databricks.labs.community_connector.interface import LakeflowConnect
+from pyspark.sql import SparkSession
+
 from databricks.labs.community_connector.sparkpds.lakeflow_datasource import LakeflowSource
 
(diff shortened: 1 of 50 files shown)


src/databricks/labs/community_connector/sparkpds/registry.py
```diff
@@ -1,171 +1,144 @@
...
+        raise ValueError(
+            f"Source '{source_name}' not found. "
+            f"Make sure the package '{package_fqn}' is installed."
+        ) from exc
+
+    for _, obj in inspect.getmembers(package, inspect.isclass):
+        if (
+            issubclass(obj, LakeflowSource)
+            and obj is not LakeflowSource
+            and getattr(obj, "__module__", "").startswith(package_fqn)
```
yyoli-db: Latent fragility (not a live bug): `startswith(package_fqn)` lacks a dot boundary. `find_data_source("github")` builds `package_fqn="...sources.github"`; a future source named e.g. `github_enterprise` would have `__module__` starting with `...sources.github` too, so the wrong source's class could match. No collision among the 22 current sources, so this is latent. Consider `== package_fqn or .startswith(package_fqn + ".")`.

yyoli-db:
tested with community connector e2e and the python pkg through a spark cluster

PR #230, opened 2026-06-26
Fix masked error reporting in HubSpot connector

wesley84:
## What

Fixes masked error reporting in the HubSpot connector's `_get_object_properties`.

Two problems on the error path:

1. **Non-f-string raise** (`hubspot.py:340`): `raise Exception("API error: {resp.status_code} {resp.text}")` — missing the `f` prefix, so the real status code and body never appeared; the message was a literal template.
2. **Swallowed exception** (`hubspot.py:343-344`): the `except` returned `{"error": ...}` (a dict). But both callers (`:267-268`, `:290`) iterate the result as a list of property dicts (`[prop["name"] for prop in properties]`), so a genuine API error surfaced as an unrelated `TypeError: string indices must be integers` far downstream — the actual failure was lost.

## Fix

- Make the message an f-string that includes the real status and body.
- Re-raise with context (`raise RuntimeError(...) from e`) instead of returning a sentinel dict, so failures surface where they happen.
- Regenerated the merged `_generated_hubspot_python_source.py`.

## Testing

Simulate-mode suite: **9 passed, 4 skipped** (`pytest tests/unit/sources/hubspot/`). Only the error path changed; the success path (`return resp.json()`) is untouched.

## How it was found

Surfaced by an AI connector-diagnosis tool pointed at this connector (it had never seen it before) — flagged the masked-error pattern by reading the source.

---

### DIFF
--- a/src/databricks/labs/community_connector/sources/hubspot/hubspot.py
+++ b/src/databricks/labs/community_connector/sources/hubspot/hubspot.py
@@ -1,8 +1,7 @@
 import json
-import random
 import time
 from datetime import datetime, timezone
-from typing import Any, Dict, Iterator, List, Tuple
+from typing import Dict, Iterator, List, Tuple
 
 import requests
 from pyspark.sql.types import (
--- a/src/databricks/labs/community_connector/sources/hubspot/hubspot.py
+++ b/src/databricks/labs/community_connector/sources/hubspot/hubspot.py
@@ -1,10 +1,20 @@
-import requests
 import json
-from pyspark.sql.types import *
-from datetime import datetime, timezone
-import time
 import random
-from typing import Dict, List, Tuple, Iterator, Any
+import time
+from datetime import datetime, timezone
+from typing import Any, Dict, Iterator, List, Tuple
+
+import requests
+from pyspark.sql.types import (
+    ArrayType,
+    BooleanType,
+    DataType,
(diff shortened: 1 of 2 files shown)


src/databricks/labs/community_connector/sources/hubspot/hubspot.py
```diff
@@ -1,10 +1,20 @@
-import requests
 import json
-from pyspark.sql.types import *
-from datetime import datetime, timezone
-import time
 import random
-from typing import Dict, List, Tuple, Iterator, Any
+import time
+from datetime import datetime, timezone
+from typing import Any, Dict, Iterator, List, Tuple
```
yyoli-db: I think this introduces some unused imports

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/hubspot/hubspot.py
wesley84: Good catch — fixed in d2b9971. Removed `random` and `typing.Any` (both were pre-existing-unused; the import reorder just surfaced them on that line). Checked every remaining import is referenced. Merged source regenerated, simulate suite green (10 passed), pylint 10/10.

yyoli-db (approved):

PR #231, opened 2026-06-26
Add OData v4 community connector

Leone67:
## Summary

Generic OData v4 community connector for Lakeflow Connect. Validated end-to-end against the live Northwind reference service and Microsoft Graph metadata.

**Read modes**
- **Snapshot** — no `cursor_field`.
- **Cursor-based incremental** — `cursor_field` set; `gt` filter, boundary trim for low-cardinality cursors, primary-key tiebreaker, no wall-clock ceiling. Every server-paged fetch carries a stable, unique `$orderby` (cursor + PK, or PK-only) so value-based skiptokens can't silently drop or duplicate rows across pages. An overlap re-read window (`cursor_lookback_seconds`, `auto` by default) re-scans a small trailing window behind the committed watermark so rows that arrive **during** a non-atomic walk — landing just below the walk's final max — are recovered on the next progressing batch (MERGE-deduped at the destination; with `cursor_lookback_dedup` — on by default — unchanged overlap re-reads are dropped connector-side before emit, so only genuine changes reach the MERGE). The `auto` window self-sizes from measured walk durations × a safety factor, down to nanosecond precision, so even fast/small sources get a non-zero overlap; the committed watermark is never floored.
- **Server-driven delta tracking** — `delta_tracking={auto,enabled}`; `Prefer: odata.track-changes` with `@odata.deltaLink` resume and in-band `_deleted` tombstones (microsoft_teams precedent).
- **Contained navigation properties** — `Parent__Child[__Grandchild...]` paths. Strategies per table:
  - N+1 (`expand_contained=false`, default) — walks parents one at a time, applies leaf filters, commits at every parent-walk boundary.
  - Single `$expand` (`expand_contained=true`) — one big nested request; rows flattened in-process. Per-fetch resume via a **depth-first resumable stack machine**: the drain walks one root→leaf path at a time, checks the `max_records_per_batch` cap between rows/fetches **at any depth**, and parks only the boundary path from root to current leaf — an **O(depth) offset regardless of fan-out width** (a breadth-first queue would park one continuation per truncated inner collection). Resume is churn-safe via chronological boundary order-keys, never positional `$skip`; in `nextlink` mode a mid-way inline collection parks as a from-start refetch whose boundary elides the already-emitted prefix. Cap deviation per batch is bounded by one HTTP response's worth of leaf rows (≤ `page_size`) in every pagination mode. Deep `$expand` continuations budget their `$top` across only the remaining collection levels (root-down), not the fixed-key ancestors, so nested pages aren't pinned to the per-level floor.
  - Incremental acceleration (`cursor_probe`, default `auto`) — the `expand_contained=false` counterpart for **deep paths with sparse changes**, with a capability **cascade**. `auto` (default) uses the nested-`$expand` change-probe where it pays off *and* the server is verified to honour `$orderby`/`$top` inside `$expand`; otherwise it falls back to an OData **`$batch`** hydrate (where the server supports `$batch`); otherwise the plain N+1 walk — **never raising** on a capability shortfall. The **change-probe**: for a leaf-owned cursor read whose leaf-parent is itself cursor-bearing (distance from the nearest snapshot ancestor > 1), one shallow `$expand(<leaf>($orderby=cursor desc;$top=1;$select=cursor))` probe per leaf-grandparent identifies which leaf-parents changed, then only those are hydrated via the normal N+1 walk. A one-time behavioural preflight verifies the server orders inner `$expand` correctly; under `cursor_probe=nested-expand` (strict) a failure **raises**, under `auto` it **cascades**. The **`$batch` hydrate** (`cursor_probe=batch`, or the `auto` fallback) issues the same per-leaf-parent `cursor gt since` reads but packs them into `$batch` requests (1000 ops/request by default, tunable via `batch:<N>`; server-driven paging, `@odata.nextLink` follow-up) — collapsing *M* round-trips into `ceil(M/N)` using only top-level single-column filters, so it's **drop-safe on servers (e.g. Hexagon Smart API) that reject nested-`$expand` options**. If the server rejects a batch for carrying too many sub-requests (matched across phrasings — “contains too many parts” / “$batch exceeds the maximum of N operations”), the working size shrinks 25% and retries (up to 10 times — enough to converge from 1000 below a ~100-part server cap) before degrading to per-parent GETs; the discovered size persists as `batch_size_ok`. The probe's dirty-parent hydrate also goes through `$batch` when supported (plain N+1 otherwise, or when an explicit `contained_fetch=single`/`1` opts out). Both emit rows identical to the plain walk; the `$batch` verdicts (`batch_ok`/`batch_size_ok`) and probe verdict (`cursor_probe_ok`) persist in the offset — only **definitive** preflight outcomes are persisted (a transient blip degrades one batch and re-probes next), and pinned non-`auto` modes scrub their verdicts so re-selecting `auto` re-runs the preflights.
- **Partitioned reads** — `SupportsPartitionedStream` for contained snapshot paths. Top-level parents are enumerated once, bin-packed across `num_partitions`, each Spark task walks only its slice.

**Auth modes**
`bearer`, `basic`, `api_key`, plus **UC-managed OAuth** via the COMMUNITY connection (`community_oauth_flow=m2m`/`u2m`): the connection layer runs the flow, refreshes tokens server-side, and injects a fresh `access_token` at query time, which the connector treats as an opaque bearer credential — the connector contains no OAuth code. The spec declares an `oauth` auth method (`client_id`/`client_secret`/`token_endpoint`, optional `oauth_scope`; per-connection `token_endpoint` since OData is generic, no provider URLs are hard-coded) and never lists the runtime-injected `access_token`/`refresh_token`. The old connector-side `auth_type=oauth2` is retired and raises a curated migration error. 401/403 surface as per-auth-mode `PermissionError` with the response body echoed truncated; with an injected token they are terminal for the query and name the connection layer as the refresh owner.

**Multi-tenant**
`SupportsNamespaces` walks multi-`<Schema>` CSDL documents. `BaseType` inheritance chain is walked when resolving `<Key>` and `<Property>` declarations, required for Microsoft Graph, Dataverse, and SAP S/4HANA Cloud where abstract base types define identity.

**Tuning knobs**
- `pagination` — page-walk strategy: `auto` (default; follow `@odata.nextLink` when emitted, else fall back to a keyset/skip seek and drain until empty), `nextlink` (strictly spec-compliant; the `$top`-free snapshot scan), `keyset`, `skip`. Under `auto` a short link-less page is **never** treated as exhaustion — the walk keeps seeking until an empty page, so a server that page-limits below the requested `$top` while suppressing `@odata.nextLink` is still read in full (a `page_next_url == url` no-progress guard bounds a server that ignores the seek; any re-fetched row is MERGE-deduped at the destination). This drain is the single, unconditional behaviour across snapshot, flat-cursor, contained leaf/ancestor, and nested `$expand` continuation walks. **`$top` is a TOTAL-result limit in OData (§11.2.5.3), not a page size**: a spec-compliant server may propagate the remaining budget through its skiptoken `@odata.nextLink`s (e.g. Northwind: `$top=1000` → page 1's link carries `$top=500` → after 1000 rows no further link), so the link chain self-terminates at the `$top` sent — which would silently cap any table larger than `page_size` at `page_size` rows. `auto` guards this: when the chain ends at exactly the budget (`fetched >= top`), it does **not** trust the short final page but seeks past the budget and keeps draining (verified end-to-end against live Northwind: `Order_Details`/`Invoices`/`Order_Details_Extendeds`, 2155 rows each, previously truncated to 1000).
- `page_size` — opt-in `$top`. **Unset → no `$top` for snapshot ingest under `nextlink`** (the server picks its page size; avoids servers that reject an explicit `$top` on a full scan); **1000 for cursor/delta ingest, and for every mode under `auto`** (which needs a `$top` to detect a page-limited response). When set, distributed triangularly across nested `$expand` levels so the cross-product stays within the budget (per-level floor of 5; remaining budget redistributes when a level clamps).
- `max_records_per_batch` — per-call leaf-row cap (default 10000). Under the framework batch reader (`LakeflowBatchReader`, `start_offset=None`) the returned offset is discarded, so the cap is always disabled there (with a `WARNING` when a user sets one) and the cursor read paths stream lazily to keep peak memory bounded to one response. Streaming readers keep the cap and resume state.
- `cursor_probe` — incremental leaf-cursor acceleration; see **Contained navigation** above. Values: `auto` (default; cascade nested-`$expand` probe → `$batch` → plain walk, never raises on capability), `nested-expand` (strict probe; raises on inner-`$expand` mis-order or an ineligible table; dirty parents hydrate via `$batch` when supported), `batch` / `batch:<N>` (skip the probe, hydrate via `$batch` at `N` ops/request — default 1000 — falling back to the plain walk if `$batch` unsupported), `false` (plain N+1 walk).
- `contained_fetch` — how the **un-cursored full** contained walks hydrate each leaf-parent (the snapshot read, and the framework batch-reader stream / `start_offset=None`). `auto` (default) packs the per-leaf-parent GETs into OData `$batch` (1000 ops/request, server-driven paging, `@odata.nextLink` follow-up), collapsing *M* round-trips into `ceil(M/1000)`; a one-shot capability preflight gates it and it transparently falls back to `single` on servers without `$batch`. `batch` is the same hydrate but **strict** (a preflight failure raises); `auto:<N>` / `batch:<N>` tune the chunk size; `single` (or `1`) is one GET per leaf-parent (the original behaviour); a bare integer `N>1` is strict like `batch:<N>`. Adaptive sizing shrinks the chunk 25% on a server “too many parts” rejection (up to 10 retries) and persists the discovered size (`batch_size_ok`). Mostly independent of `cursor_probe` (which governs the *incremental* read) — but an explicit `single`/`1` also forces the incremental hydrate down the plain N+1 walk (except under an explicit `cursor_probe=batch`, which wins the conflict). Identical rows either way.
- `cursor_lookback_seconds` / `cursor_lookback_factor` / `cursor_lookback_max_seconds` — overlap re-read window behind the committed watermark for non-atomic walks. `auto` (default) self-sizes from the max of recent measured walk durations × factor (1.5, configurable), clamped to a ceiling (3600 s, configurable), at nanosecond precision; an explicit integer fixes the window; `off`/`0` disables it. No-op for non-timestamp cursors.
- `cursor_lookback_dedup` — suppresses the redundant re-upserts the overlap window produces: an exact, capped `{composite-PK → [cursor, content-hash]}` seen-set rides the offset; a re-fetched row matching key **and** whole-row hash was already delivered unchanged and is dropped before emit. Whole-row hashing means a source update that doesn't advance the cursor still delivers. `on` by default (5000-entry cap); `off` restores the blind re-emit; a positive integer sets the cap. Every failure direction (cap overflow, mid-cycle resume shrink, corrupt checkpoint state, non-timestamp cursors) degrades to MERGE-idempotent re-emits — never suppression of an undelivered row (which is why a probabilistic set is ruled out). Genuine in-window changes now deliver promptly even on quiescent triggers via the seen-set's offset delta.
- `cursor_nulls` — null cursor policy: `coalesce` (default; substitute a deterministic synthetic floor for the watermark, configurable year via `coalesce:<YYYY>`, emit the real null), `error` (raise on an all-null batch), `ignore` (drop null-cursor rows).
- `filter_at_<segment>` / `filter_at_<idx>` — per-segment `$filter` for contained paths.
- `exclude_ancestor_columns` — comma-separated list of synthetic ancestor-FK column names (`<segment>_<pk>`) to drop from a contained table's destination, or `*` to drop them all. Applied at the single FK-resolution source of truth, so excluded columns leave the schema, the composite primary key, and the stamped rows together. **Only synthetic ancestor-FK columns can be dropped** — a real leaf/own table column named here is kept (and warned about), so the option can never remove an actual source column. No effect on flat tables.
- `timeout_seconds`, `max_retries`, `retry_max_delay_seconds`, `metadata_cache_ttl_seconds`, `extra_headers`, `verbose_http_logging` (+ `verbose_http_log_body_chars`).

**Error UX**
- Transient HTTP 429/500/502/503/504 retried with `Retry-After` + capped exponential backoff; network errors retried; bad JSON on a 200 retried with the body surfaced.
- Cursor-mode no-progress (offset can't advance) → `RuntimeError` instead of a silent infinite loop / data loss.
- A *complete* same-cursor leaf cohort larger than the cap is emitted in full and the walk continues (no spurious `RuntimeError`).
- `cursor_probe` against a server that mis-orders inner `$expand` → under `nested-expand` (strict) fail fast with an actionable error; under the default `auto` cascade to `$batch`/the plain walk (never a silent row drop, never a hard failure). The `$batch` capability preflight is single-attempt (fail-fast), so a server without `$batch` doesn't stall on the retry/backoff loop — and only **definitive** verdicts persist in the offset: a transient failure (network blip, 5xx, auth hiccup) degrades that one batch to the plain walk and re-probes the next, never permanently pinning the stream to the slow path. A resumed truncated walk that completes with nothing left to emit clears its parked checkpoint instead of freezing at `parent_idx` forever.
- Service-root-relative `@odata.nextLink`s resolved without path doubling.
- Delta-tracking 410 (token expired) → silent re-bootstrap; sparse-property delta response → `RuntimeError` (refuses null-overwrite at the destination).
- Server depth caps → HTTP 4xx surface verbatim; no client-side fallback.

## Test plan

- [x] **Unit + simulator** — 631 passed, 12 skipped offline (`tests/unit/sources/odata/test_odata_lakeflow_connect.py`). The 12 skips are the `SupportsPartitionedStreamTests` partitioned-table contract tests — the flat Northwind simulator corpus has no partitionable table, so contained partitioning is covered by 11 bespoke `test_partition_*` tests. Includes regressions for the depth-first expand-drain park/resume (O(depth) offset bound; nextlink mid-inline from-start park churn-safety), the `$top`-as-total-limit drain (Northwind shape), the `cursor_probe` `$batch` cascade (hydrate, `@odata.nextLink` continuation, 405→plain-walk fallback, `auto`-cascade-on-mis-order), the `contained_fetch` grammar (`auto`/`auto:<N>`/`batch:<N>`/bare-`N` strictness, snapshot + batch-reader-stream hydrate, fallback, invalid-value), adaptive `$batch` sizing (shrink-and-retry, give-up sentinel, seeded `batch_size_ok`, converge-below-a-100-part-cap), definitive-vs-transient preflight verdicts, the final-chunk truncation-resume checkpoint clear, and the verdict-scrub ownership rules. Also the `cursor_lookback_dedup` suite: suppression, changed-row-same-cursor re-delivery (the PK+cursor-keying hazard), deterministic cap eviction (PK tiebreak; non-timestamp cursors never crash), corrupt-checkpoint tolerance, auto-lookback history non-pollution on quiescent delta batches, and the on-by-default parse grammar.
- [x] **Live record-mode validation** — refreshed 2026-07-08 (`CONNECTOR_TEST_MODE=live`) against `https://services.odata.org/V4/Northwind/Northwind.svc/`. 3 endpoints validated, **0 mismatched / 0 drift**.
- [x] **End-to-end row + value validation (deep contained, live)** — full pipeline ingest of a 5-level contained path (`Instances → Projects → WorkPackageDetails → WorkPackagePunchlists → WorkPackages`) compared source-vs-destination by composite primary key: **9,715 rows across all 5 tables, 0 missing / 0 extra / 0 value mismatches** over up to 102 columns each (strings, decimals, timestamps incl. `RecordLastModified`). Confirms completeness and value fidelity, not just counts; transient WorkPackages lag during the source's growth phase was proven to be commit-cadence lag (every missing key landed on the next commit), never loss.
- [x] **Live metadata validation against Microsoft Graph** — `$metadata` parses; entity sets discovered; schemas + PKs resolve via `BaseType` chain walk.
- [x] **Pylint** — 10.00/10 against the CI flag set.
- [x] **Documentation audit** — README cross-checked line-by-line against the implementation (auth wiring, per-table options/defaults, delta tracking, contained-nav mechanics, multi-tenant, limitations). Discrepancies corrected: auth auto-inference is `bearer`-only (other modes need explicit `auth_type`); `cursor_nulls=coalesce` default ingests null-cursor rows rather than raising; retry set is 429/500/502/503/504; `guid` cursors are synthesisable. Stale in-code comments fixed too. The per-table options reference was restructured into a quick-reference table + per-option subsections for readability (content-preserving).
- [x] **Self-review** — `/self-review-connector` re-scored 98/100 (READY) at the current head; the full report is the sticky self-review comment on this PR.
- [x] **Bundle freshness** — `_generated_odata_python_source.py` regenerated from current sources.

This PR is assisted by Isaac Autopilot: [View the full session spec, test plan, and artifacts](http://go/isaac-autopilot/cli-viewer/1781843243671/yong-li_data).

---
**Update (2026-07-23):** Fixed pylint `R0915 too-many-statements` in `_read_table_dispatch` (extracted `_init_read_scoped_state`) so the `Pylint / build (3.10)` required check passes. Pure refactor — no wire behavior change; confirmed by a fresh live record-mode run against the public Northwind service (0 drift) and the full simulate suite (618 passed / 12 skipped).


Leone67:
# Self-Review — odata

**Overall: 99 / 100 — READY**

Run at: 2026-07-23T10:37:00Z (post pylint fix + record-mode refresh)

Both prior findings are resolved. A11 pylint `R0915 too-many-statements` is
fixed (extracted `_init_read_scoped_state` from `_read_table_dispatch`; pylint
10.00/10, exit 0 under the exact CI invocation). B11 staleness is cleared by a
fresh live record-mode run against the public read-only Northwind service —
0 mismatched, validation report now newer than the connector source, which also
independently confirms the refactor changed no wire behavior. 617 tests pass
live / 618 simulate. The sole remaining non-pass is the deliberate, documented
A8 `IntegerType` mapping (MINOR warn, unchanged per developer request).

## Top recommendations

1. **MINOR** — `Edm.Byte/SByte/Int16/Int32` map to `IntegerType` rather than
   `LongType` — `odata.py:163-166`. Deliberate and documented (framework
   `parse_value` lacks Short/Byte support). No action needed.
2. **INFO** — Live validation shows `endpoints_missing_spec: 34`: the real
   Northwind service exposes ~34 entity sets the simulator intentionally does
   not model (it models only `Customers`/`Orders`). Expected for a generic
   connector; all 3 spec endpoints matched with 0 drift.

## A. Connector implementation — 23 / 27
- ✅ A1. Class extends LakeflowConnect — `odata.py:1151`
- ✅ A2. All abstract methods implemented — `odata.py:1276, 1324, 1429, 1476`
- ✅ A3. No `cdc_with_deletes` tables → `read_table_deletes` not required — `odata.py:2379`
- ✅ A4. Read pattern deliberate: snapshot / CDC cursor stream / partitioned
       stream (`PartitionMixin`) + contained N+1 — `odata.py:1563, _partition.py:92`
- ✅ A5. Termination contract is deliberate: no `_init_time` wall-clock cap;
       instead `max_records_per_batch` + no-progress guard over a finite source
       (`_finalize_cursor_read`, pagination guard `odata.py:3339`). Documented.
- ✅ A6. `max_records_per_batch` honored on incremental tables; append-only /
       snapshot streams uncap server-side (`_BATCH_UNCAPPED`) not client-truncate — `odata.py:1687`
- ✅ A7. Single HTTP call site, carries `timeout=self.timeout` — `odata.py:3961`
- ⚠️ A8. `IntegerType` used for `Edm.Byte/SByte/Int16/Int32` — `odata.py:163-166`
       (deliberate, documented; see Top rec #2)
- ✅ A9. `get_table_schema` returns `StructType`, does not flatten nested fields
- ✅ A10. Imports clean — only interface/libs/sources-odata/requests/pyspark/std-lib — `odata.py:54-148`
- ✅ A11. **Pylint clean** — 10.00/10, exit 0 under CI flag set (was `R0915 51/50`;
       fixed by extracting `_init_read_scoped_state` — `odata.py:1563`)
- ✅ A12. No ingestion-agent surface — none found
- ✅ A13. `ODataDataSource(LakeflowSource)` binds `_lakeflow_connect_cls` — `__init__.py:12-13`

- ✅ B1. Test file exists — `test_odata_lakeflow_connect.py`
- ✅ B2. Subclasses `LakeflowConnectTests` + `SupportsPartitionedStreamTests` — `:34`
- ✅ B3. `connector_class = ODataLakeflowConnect` — `:55`
- ✅ B4. `simulator_source = "odata"` — `:56`
- ✅ B5. `replay_config` set; keys (`service_url`, `auth_type`, `token`) match `__init__` reads — `:61-65`
- ✅ B6. `endpoints.yaml` exists
- ✅ B7. Corpus JSON present — `corpus/Customers.json`, `corpus/Orders.json`
- ✅ B8. Connector hits a single dynamic entity-set URL; 2 spec entries + `$metadata` cover it
- ✅ B9. `test_read_terminates` passes (simulate mode)
- ✅ B10. All tests pass — 618 passed / 12 skipped simulate (34.2s); 617 passed / 13 skipped live (64s)
- ✅ B11. Fresh live record run 2026-07-23 10:36 (public Northwind service),
       newer than `odata.py`/`endpoints.yaml`; summary 3/3 validated,
       **0 mismatched** (34 missing_spec — see Top rec #2). Confirms the pylint
       refactor changed no wire behavior.
- ✅ B12. All spec endpoints hit (`$metadata`, `Customers`, `Orders`)
- ✅ B13. `synthesize_future_records: {cursor_field: OrderDate, count: 3}` on Orders — `endpoints.yaml:39`
- ✅ B14. No `list_insertable_tables` → no write-back → `test_utils_class` correctly absent

## C. Artifacts — 28 / 28
- ✅ C1. `odata.py` compiles clean (`py_compile`)
- ✅ C2. `odata_api_doc.md` non-empty, documents entity-set tables
- ✅ C3. `connector_spec.yaml` parses; has params + `external_options_allowlist`;
       `oauth.flow: m2m`, `client_id`/`client_secret` listed, no runtime tokens as params
- ✅ C4. `README.md` non-empty, mentions tables + connection parameters
- ✅ C5. `pyproject.toml` parses; `dependencies` includes `requests`
- ✅ C6. `_generated_odata_python_source.py` (10:31) newer than `odata.py` (10:30) — regenerated post-fix
- ✅ C8. Corpus files are valid JSON
- ✅ C9. Test class present (see B1)
- ✅ C10. No runtime `pyspark` — kept in `[optional-dependencies].dev` only

## D. Security — 22 / 22
- ✅ D1. No hardcoded secrets
- ✅ D2. No dynamic `eval`/`exec`/`compile`/`__import__` (matches are `re.compile`)
- ✅ D3. No `subprocess`/`os.system`/`shell=True`
- ✅ D4. No `verify=False`
- ✅ D5. `pickle.load` (`odata.py:4355`) is on the connector's OWN `$metadata`
       cache, gated by uid-ownership + shape check — not response bodies
- ✅ D6. No unsafe `yaml.load`
- ✅ D7. No path traversal — cache path derived from `service_url` hash
- ✅ D8. No secrets in logs; embedded-userinfo `service_url` rejected at construction — `odata.py:1180`
- ✅ D9. `http://` only in XML namespace URIs — `odata.py:180-181`
- ✅ D10. Deps bounded (`requests>=2.31.0,<3.0`, `lakeflow-community-connectors>=0.1.0,<1.0`)

- ✅ E1. Tables dynamic (from `$metadata`); API doc + README use consistent Northwind examples
- ✅ E2. Connection-parameter keys agree across `__init__` / spec / README;
       UC-injected `access_token` correctly exempt
- ✅ E3. Schema columns derive from `$metadata`; subset of API doc fields
- ✅ E4. Primary keys resolved from CSDL `<Key>` — `odata.py:1437`
- ✅ E5. `external_options_allowlist` covers every `table_options` key read
- ✅ E6. README quick-start uses spec parameter keys
- ✅ E7. README does not contradict API doc on auth/pagination/rate limits

src/databricks/labs/community_connector/sources/odata/_partition.py
```diff
@@ -0,0 +1,787 @@
...
+    """Split ``rows`` into ``num_partitions`` partition descriptors.
+
+    Each partition carries a contiguous slice — keeps cursor ordering
+    stable within a partition (the ``_discover_top_parent_rows`` caller
+    sorts by cursor when one is set). Empty bins are dropped so the
+    framework doesn't spawn no-op executors.
+    """
+    if not rows:
+        return []
+    bin_size = max(1, (len(rows) + num_partitions - 1) // num_partitions)
```
yyoli-db: nit: (parallelism, not correctness): `bin_size = ceil(n / num_partitions)` with contiguous slicing produces `ceil(n / ceil(n/p))` bins, which is fewer than `num_partitions` for many inputs (n=9, p=4 → bin_size 3 → 3 bins; n=10, p=6 → 5 bins). Coverage is exactly-once with no drop/overlap (verified), so no data issue — just fewer executors than the user requested. A round-robin split (or `array_split`-style) would hit the requested partition count.

src/databricks/labs/community_connector/sources/odata/_contained.py
```diff
@@ -0,0 +1,6086 @@
...
+                    # raise is not ``_BatchTooManyParts`` — so a pass verdict
+                    # for an id-less server would pin ``batch_ok: true`` and
+                    # then fail EVERY hydrate with "missing sub-response id"
+                    # instead of degrading to plain GETs.
+                    # Last-wins on duplicate ids, matching how _post_batch's
+                    # by_id dict resolves them — probe and hydrate must judge
+                    # the same sub-response.
+                    by_id = {str(r.get("id")): r for r in subs if isinstance(r, dict)}
+                    sub = by_id.get("0")
+                    sub_status = int(sub.get("status", 0) or 0) if sub else None
```
yyoli-db: nit: a *missing* sub-response `status` maps to `0` here (`int(sub.get("status", 0) or 0)`), and `0 < 400` then reads as a definitive pass and persists `batch_ok=True`. It self-heals — the real hydrate (`_checked_batch_subresponse`) treats `status is None` as bad and re-issues a plain GET, so no data loss — but a malformed-but-2xx `$batch` envelope optimistically pins `batch_ok=True`, unlike the "only definitive outcomes are cached" discipline used elsewhere. Consider recording nothing when the probe sub-response carries no status.

src/databricks/labs/community_connector/sources/odata/_contained.py
```diff
@@ -0,0 +1,6086 @@
...
+        the boundary trim and (via the caller) the watermark — the
+        ``cursor_nulls`` resolver, so a null cursor can resolve to a
+        synthetic floor without mutating the emitted row. ``skip_null``
+        drops rows with a real null cursor (``cursor_nulls=ignore``)."""
+        if effective is None:
+
+            def effective(row):
+                return row.get(cursor_field)
+
+        emitted: list[dict] = []
```
yyoli-db, line 4876: qq: rows inside the active lookback overlap (cursor `<= since`) are appended to `emitted` but never advance `countable`, so `max_records_per_batch` doesn't bound them. A table with a large number of rows inside the active lookback window will materialize the whole in-window set in memory regardless of the cap. This is design-acknowledged and `_apply_lookback_dedup` suppresses the re-emits downstream — but peak memory scales with in-window row count, not the cap. Is that acceptable for expected window sizes, or should the overlap re-read also page-stream?

src/databricks/labs/community_connector/sources/odata/_helpers.py
```diff
@@ -0,0 +1,494 @@
...
+            # data pending (a flat False here is duplicate-safe for the
+            # watermark fold but NOT for the re-filter). Non-numeric
+            # incomparable pairs keep the arbitrary-but-consistent False,
+            # matching ``_chain_strictly_before``'s documented posture:
+            # degrade duplicate-safe, never raise out of a fold.
+            num_a, num_b = _as_exact_number(a), _as_exact_number(b)
+            if num_a is not None and num_b is not None:
+                return num_a > num_b
+            return False
+    if a == b:
```
yyoli-db, line 187: Nit: the numeric-string fraction tie-break makes ordering asymmetric for numerically-equal but textually-different values — e.g. `cursor_newer("5000.0", "5000")` is `True` while the reverse is `False`. Both directions are duplicate-safe (same instant), so no data impact; just flagging that it isn't a strict antisymmetric order for that pair class.

yyoli-db (changes requested):
left some small comments and please file a ticket to get LPP approval
see http://go/community-connectors/tracking

yyoli-db:
@Leone67 also please ping me when you make the updates through slack. ty!

src/databricks/labs/community_connector/sources/odata/_partition.py
Leone67: Fixed in e46dded: balanced `divmod` split — the first `n % bins` slices carry one extra row, so every requested partition materializes (`min(num_partitions, n)` non-empty bins; n=9, p=4 → `[3,2,2,2]`), still contiguous so per-partition cursor ordering is preserved. Regression: `test_bin_pack_hits_requested_partition_count` (bin count, sizes, exactly-once coverage, no empty bins).

src/databricks/labs/community_connector/sources/odata/_contained.py
Leone67: Fixed in e46dded: `sub.get("status")` with no default — a missing `status` now flows to the existing "not a consumable envelope" branch (definitive FAIL, `batch_ok` pinned false), matching the id-echo discipline four lines up, so the hydrate goes straight to plain GETs instead of paying a doomed `$batch` POST every read. Regression: `test_batch_probe_missing_substatus_is_definitive_fail`.

src/databricks/labs/community_connector/sources/odata/_contained.py
Leone67: Good question — answered with a fix rather than a rationale (e46dded): `cursor_lookback_dedup` (on by default) now filters **at emit time** (`_LookbackDedupFilter`, armed per cursor read, consulted at all four walk emit points). A proven-unchanged overlap row never materializes in the batch buffer — only its ~100-byte seen-entry is carried — so peak memory scales with rows actually delivered (genuine changes + new rows, the irreducible floor since the offset needs a full batch scan) plus entry bookkeeping, not with window churn. Entries for delivered rows are computed in finalize from the post-trim buffer, so an admitted-then-trimmed row can never leave an entry that would suppress its future re-fetch. The 10 pre-existing dedup tests now run through the streaming path unchanged (semantic equivalence), plus `test_lookback_dedup_filter_streams_suppression` for the filter contract.

Why not page the overlap against the cap instead: overlap rows are below the watermark, so counting them means every call commits park-state "progress" and a quiescent source can never idle — an eternal window re-read per trigger. A paged design is only sound on top of the seen-set's committed entries (they're what let a paged cycle converge); the README documents that as future work for the all-churn-window case, and documents the `off`-mode bound (memory tracks delivered rows by contract there — size via `cursor_lookback_max_seconds`).

src/databricks/labs/community_connector/sources/odata/_helpers.py
Leone67: Fixed in e46dded: exact-Decimal bridge in the key-tie path — numerically-equal renderings ("5000.0" vs "5000") now read same-instant in **both** directions, restoring antisymmetry (and preventing watermark-text flap on a rendering-alternating server). Bonus: Int64 cursors beyond 2^53 (whose float sort keys collapse) now order truly via the same bridge. ISO texts never Decimal-parse, so the sub-microsecond fraction tie-break is untouched. Regression: `test_cursor_newer_numeric_rendering_symmetry`.

src/databricks/labs/community_connector/sources/odata/connector_spec.yaml
```diff
@@ -0,0 +1,185 @@
...
+          description: API key value.
+        - name: api_key_header
+          type: string
+          required: false
+          description: Header name to carry the API key. Default ``x-api-key``.
+
+    - name: oauth2
+      description: >
+        OAuth 2.0. Two grant flows are supported via the same parameter
+        set:
```
yyoli-db: I did not pay attention to this piece in previous round. 
The `COMMUNITY` connection supports different oauth types automatically, meaning we don't have to deal with the token refreshing in the connector code but the UC logic will automatically handle it.

Please take a look at the gmail connector as an example. Also, the latest skills should include these patterns. Please let me know if you are using agent and the repo skills just somehow did not do the correct thing. 
Thanks for working on this!

Leone67: Thanks — adopted wholesale in 3e02cb3, and it made the connector meaningfully smaller (+339/−1,526):

- **Spec**: the `oauth2` auth-method (6 `oauth2_*` params) is replaced by a `connection.oauth` block (`flow: m2m`). One deliberate deviation from the gmail shape, since OData is a *generic* connector: no provider URLs are hard-coded — each connection supplies its own `token_endpoint` (plus `authorization_endpoint` for `u2m`) at creation time, which the CLI/UC already support as connection options. `client_id`/`client_secret`/`token_endpoint`/`oauth_scope` are declared as parameters; the runtime-injected `access_token`/`refresh_token` are not.
- **Connector**: all OAuth code removed — token minting, refresh-token grant + rotation tracking, pre-emptive expiry, the 401 re-mint path, and token-endpoint error handling. The connector now just detects the UC-injected `access_token` (no `auth_type`) and sends it as an opaque bearer, gmail-style. `auth_type=oauth2` raises a curated migration error; a mid-read 401 under an injected token names the connection layer and the retry path.
- **Tests/docs**: 21 OAuth-machinery tests retired, 4 added (injected-token auth on every request, retirement error, injected-token 401 UX, unknown-auth message); README/api-doc auth story rewritten around the COMMUNITY connection. Full suite 618 passed; live Northwind validation clean on this build.

On your skills question: we did use the repo skills, and the oauth-block pattern *is* documented there — this one's on our spec, not the skills. We initially modeled OAuth as a static-credential auth method because the generic-endpoint concern (no fixed `token_url` to put in the block) made the gmail shape look inapplicable; realizing `token_url` maps to a per-connection `token_endpoint` option resolved it. A one-line note in the skill/template that generic connectors may omit the URLs from the oauth block and document them as connection parameters would prevent the same misread by the next generic connector.

---

**Update (d474f36):** exercising this end-to-end surfaced one more spec-shape detail. The labs CLI's `create_connection` validates options via `detect_auth_method`, which only matches entries **inside** `auth_methods` — so a spec with static auth methods plus a `connection.oauth` block could never validate an OAuth create ("No valid authentication method detected"). Following the youtube spec's shape, the app-identity params (`client_id`/`client_secret`/`token_endpoint` required, `oauth_scope` optional) now live under an `oauth` entry in `auth_methods` instead of the common parameter list. Verified for real: the CLI detected the method, stamped `community_oauth_flow=m2m`, and created a working UC COMMUNITY connection against a live OData service (UC ran the client-credentials flow server-side at creation). The README also gained a concrete Python SDK example for the m2m connection create.

yyoli-db: thanks. have you run the CLI tool to test this against some real instance to validate the oauth flow here?

Leone67: fixed

src/databricks/labs/community_connector/sources/odata/connector_spec.yaml
```diff
@@ -0,0 +1,187 @@
...
+          description: >
+            OAuth app client id used by the connection layer to run the
+            flow and refresh tokens server-side.
+        - name: client_secret
+          type: string
+          required: true
+          secret: true
+          description: >
+            OAuth app client secret issued alongside ``client_id``.
+        - name: token_endpoint
```
yyoli-db, line 69: Minor (spec self-consistency): the `token_endpoint` description instructs u2m users to "also supply `authorization_endpoint`", but `authorization_endpoint` is not a declared parameter of the `oauth` method. Harmless for the CLI path (it's exempted via `OAUTH_OPTION_KEYS` in `oauth_flow.py`), but consider declaring it as an optional `oauth` parameter for spec/SDK-path consistency. Not needed for `m2m`.

src/databricks/labs/community_connector/sources/odata/connector_spec.yaml
```diff
@@ -0,0 +1,187 @@
...
+      type: string
+      required: false
+      description: >
+        Cap on the response body snippet length included in each INFO
+        log line when ``verbose_http_logging`` is on. Default 500. Raise
+        for richer diagnostics or lower to cut log volume on
+        high-throughput pipelines.
+
+# Per-table options the pipeline may pass through.
+external_options_allowlist: "namespace,cursor_field,select,filter,filter_at_*,page_size,max_records_per_batch,cursor_nulls,delta_tracking,expand_contained,num_partitions,pagination,exclude_ancestor_columns,cursor_lookback_seconds,cursor_lookback_factor,cursor_lookback_max_seconds,cursor_lookback_dedup,cursor_probe,contained_fetch"
```
yyoli-db: Major — `basic`/`api_key` auth may be unreachable through the spec-validated path. The spec declares `auth_methods` (bearer/basic/api_key/oauth) but never declares an `auth_type` connection parameter, while `odata.py:4102` needs `auth_type` to select basic vs api_key (only bearer/oauth auto-infer from `token`/`access_token`). If the UC option validator rejects undeclared params, `auth_type=basic` is rejected as unknown; and supplying only `username`/`password` falls through every branch in `_build_session` and applies NO auth (→401). Compare `fhir/connector_spec.yaml`, which declares `auth_type` as a shared parameter. Please confirm the UC-layer semantics.

yyoli-db: Major — `filter_at_*` allowlist entry relies on wildcard matching the allowlist layer may not implement. `filter_at_*` is listed in `external_options_allowlist`, but in-repo allowlist handling is exact comma-split with no glob. The connector reads concrete keys like `filter_at_Instances`/`filter_at_0` (`_contained.py:1008`). If the UC gate is exact-match, the documented per-segment filter feature is silently stripped before reaching the connector. Please confirm the UC-layer matching semantics.

src/databricks/labs/community_connector/source_simulator/specs/odata/handlers/delta_entity_set.py
```diff
@@ -0,0 +1,267 @@
+"""Simulator handler for OData v4 delta-tracking entity-set endpoints.
```
yyoli-db: this seems dead code ?

src/databricks/labs/community_connector/sources/informix/sqli.py
```diff
@@ -0,0 +1,1634 @@
...
+        )
+        cursor = 0
+        for index, column in enumerate(description.columns):
+            kind = column.type_code & 0xFF
+            start = cursor if variable_layout else column.position
+            if variable_layout and kind in {13, 16}:
+                if start >= len(payload):
+                    raise SqliProtocolError("truncated variable VARCHAR tuple")
+                limit = start + 1 + payload[start]
+            elif variable_layout and kind in {40, 41, 43, 45, 46}:
```
yyoli-db: Major — BOOLEAN (type 45) mis-framed in variable-layout tuple decoding. Type 45 (SQLBOOL) is fixed 1-byte: handled as fixed in `_fixed_result_size` ({23,45}) and excluded from the variable-layout trigger set, yet wrongly included in the length-prefixed "complex" framing branch ({40,41,**45**,46}). A SELECT returning a VARCHAR (triggers variable layout) plus a BOOLEAN frames the boolean as a 5+N-byte complex value using a bogus 4-byte length read from its payload, misaligning that and every subsequent column → `SqliProtocolError` or silently wrong values. Remove 45 (and review 41 SQLUDTFIXED, same over-broad set).

src/databricks/labs/community_connector/sources/odata/odata.py
```diff
@@ -0,0 +1,5308 @@
...
+          :class:`requests.Timeout`,
+          :class:`requests.ChunkedEncodingError`. The server didn't
+          finish sending an HTTP response (TCP reset, remote disconnect,
+          read/connect timeout, half-closed mid-body), so there's no
+          ``Retry-After`` to honour — pure exponential backoff. After
+          ``max_retries`` attempts, re-raises the original exception
+          type with the attempt count appended; ``__cause__`` preserves
+          the original traceback for triage.
+
+        Inner per-attempt logic (see ``_http_get_once``):
```
yyoli-db: Minor (doc-only) — `_http_get` docstring describes an OAuth token-refresh flow that does not exist. It claims "Pre-emptive token refresh" and "Reactive token refresh — 401 … mint a fresh token and retry once," but the code has none: `_http_get_once` raises `PermissionError` on the first 401/403 and `_http_get` never catches it. Misleads maintainers triaging 401s; the injected-token expiry limit is real and documented at `odata.py:4009`, but this docstring contradicts it.

src/databricks/labs/community_connector/sources/odata/_partition.py
```diff
@@ -0,0 +1,796 @@
...
+                self._tag_with_ancestor_fks(row, segments, chain, fk_columns)
+                if cursor_level == len(segments) - 1:
+                    # Leaf-cursor mode: filter per row by ``cursor gt
+                    # cursor_lower`` — chronological via ``_cursor_le``,
+                    # never lexical (``.5Z`` vs ``Z`` renderings invert
+                    # under string order).
+                    rec = row.get(cursor_field)
+                    if skip_null and rec is None:
+                        continue
+                    if (
```
yyoli-db, line 735: Minor — leaf-cursor re-filter can drop rows when the cursor rendering/type flips mid-stream. `_cursor_le(rec, cursor_lower)` == `not cursor_newer(...)`; for an incomparable pair (ISO-timestamp row vs numeric-string watermark), `cursor_newer` returns `False` by design, so `_cursor_le` returns `True` and the genuinely-new row is silently skipped. The numeric bridge in `_helpers.py:183` only covers numeric-vs-numeric. Needs a heterogeneous cursor column, hence Minor. (Identical serial re-filter at `_contained.py:4934`.)

src/databricks/labs/community_connector/source_simulator/specs/odata/handlers/entity_set.py
```diff
@@ -0,0 +1,391 @@
...
+  * ``$filter`` — supports ``field gt VALUE``, ``field ge VALUE``,
+    ``field lt VALUE``, ``field le VALUE``, ``field eq VALUE``, joined
+    by ``and``. Parentheses around individual conjuncts (the connector
+    emits ``(A) and (B)``) are tolerated. Date / timestamp literals
+    pass through bare; string literals are single-quoted.
+  * ``$orderby`` — comma-separated list of ``field [asc|desc]``
+    terms; subsequent terms are tie-breakers.
+  * ``$top`` — page-size cap on the response.
+  * ``$skip`` — offset into the filtered+sorted list.
+  * ``$select`` — projects the named columns (passed through unchanged
```
yyoli-db: Minor — the docstring says `$select` "projects the named columns," but `serve_entity_set` never applies it. `$select` schema behavior is only checked via `responses` mocks, so a projection bug wouldn't surface through the simulator path the contract suite uses.

src/databricks/labs/community_connector/sources/odata/pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+]
+
+[tool.setuptools]
+packages = ["databricks.labs.community_connector.sources.odata"]
+
+[tool.setuptools.package-dir]
+"databricks.labs.community_connector.sources.odata" = "."
+
+[tool.setuptools.package-data]
+"*" = ["*.yaml", "*.md"]
```
yyoli-db: Minor — package-data omits `*.json`. Uses `"*" = ["*.yaml", "*.md"]`; the established pattern (example, amplitude, …) is `["*.json", "*.md", "*.yaml"]`. Harmless today (no `*.json` in the odata source dir), but it diverges from convention and would silently drop any future JSON data file from the wheel.

src/databricks/labs/community_connector/sources/odata/odata.py
```diff
@@ -0,0 +1,5308 @@
...
+        (nesting depths and namespaces verify differently), so it seeds only
+        under the offset's own :meth:`_expand_shared_key` — a scalar here
+        would hand table A's verdict to table B and then bake it into B's
+        offset. Persisted back by :meth:`_merge_capability_caches`."""
+        off = start_offset or {}
+        if "or_filter_ok" in off:
+            self.__dict__["_or_filter_ok"] = bool(off["or_filter_ok"])
+        if "batch_ok" in off:
+            self.__dict__["_batch_supported"] = bool(off["batch_ok"])
+        if "batch_size_ok" in off:
```
yyoli-db, line 1983: Trivial — coerces checkpoint-sourced values with bare `int()` (`int(off["batch_size_ok"])`; also `:3448`), bypassing the curated parsers used elsewhere. A corrupt/hand-edited checkpoint raises an opaque `ValueError` instead of a curated message. Low likelihood (connector-written values).

src/databricks/labs/community_connector/sources/odata/odata.py
```diff
@@ -0,0 +1,5308 @@
...
+                # A single clause goes on the wire as-is. Wrapping it
+                # in parens would compound with any pre-wrapped clause
+                # passed via ``extra_filter`` (e.g. a multi-source
+                # ``combine_filters`` result), producing triple-paren
+                # ``$filter=((A) and (B))`` shapes that are harder to
+                # eyeball.
+                params.append(f"$filter={filters[0]}")
+            else:
+                params.append(f"$filter={' and '.join(f'({f})' for f in filters)}")
+        if order_by:
```
yyoli-db, line 3207: Trivial — appends `$orderby`/`$filter`/`$select` without percent-encoding, e.g. `$orderby=Id asc,Name asc` with literal spaces. Broadly tolerated and presumably tested live, but a strict gateway could 400.

.gitignore
```diff
@@ -43,6 +44,7 @@ Thumbs.db
 
 # Claude
 .claude/agent-memory-local/
+agents.md
```
yyoli-db: this seems too strict. maybe just have it in your own local repo

src/databricks/labs/community_connector/sources/informix/_generated_informix_python_source.py
yyoli-db, line 1: we should completely remove this `informix` dir from this PR

yyoli-db (changes requested):

yyoli-db (approved):

PR #232, opened 2026-06-26
Add Entra ID service-principal (m2m OAuth) auth to Azure DevOps connector

wesley84:
## What

Adds Microsoft Entra ID **service-principal** authentication (OAuth 2.0
client-credentials / m2m) to the Azure DevOps connector, alongside the existing
PAT auth. Driven by an enterprise ask (customer wants service-principal OAuth
instead of per-user PATs).

The connector now supports two auth methods, declared under `auth_methods` in
`connector_spec.yaml`

## Why

PATs are tied to a user and are increasingly restricted by enterprise security policies. Service-principal OAuth is the governed, app-identity path many organizations require.

## How

- **Connector (`azure_devops.py`)**: `_configure_auth()` selects the scheme from
  the connection options it receives. If UC injected an `access_token`, it uses
  `Authorization: Bearer`; otherwise it falls back to `personal_access_token`
  (HTTP Basic). `access_token` takes precedence when both are present. The
  previous hard requirement on `personal_access_token` is now a combined check:
  the connector requires *either* `access_token` (service_principal / OAuth) or
  `personal_access_token` (pat). No call sites changed. Note: the connector does
  **not** refresh the token — token lifetime/refresh is owned entirely by UC.
- **Spec (`connector_spec.yaml`)**: restructured to the `auth_methods` shape
  with `pat` and `service_principal` methods plus shared parameters
  (`organization`, `project`). Adds an `oauth` block with `flow: m2m`,
  `token_url: https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token`,
  and `scopes: 499b84ac-1321-427f-aa17-267ca6975798/.default` (the first-party
  Azure DevOps resource `.default` scope).
- **CLI (`cli.py`)**: adds `_interpolate_oauth_placeholders()`, which resolves
  `{param}` placeholders in resolved OAuth connection options from the
  user-supplied `--options`. This lets the per-tenant `token_url` be filled from
  the `tenant_id` option at connection-creation time. A missing/empty referenced
  option is a hard error rather than leaving a literal `{tenant_id}` in the
  stored connection option.
- **README**: documents both auth methods and the service-principal setup steps.
  Note: `service_principal` (m2m) is **not yet supported through the
  community-connector UI** ("Add data" flow) — create the connection with the
  `community-connector` CLI for now. The `pat` method works in both the UI and
  the CLI.

## Testing

`pytest tests/unit/sources/azure_devops/` → **34 passed, 2 skipped** (offline). New tests cover: auth-mode branching (pat default, oauth selection, conditional-required validation, unknown mode) and the token flow (acquisition, caching/refresh, request-failure). pylint **10.00/10** on the changed source. Merged source regenerated.

> Live verification against a real Entra SP + Azure DevOps org is pending (needs SP credentials); the auth-branch and token logic are covered offline.

---

### DIFF
--- a/src/databricks/labs/community_connector/sources/azure_devops/README.md
+++ b/src/databricks/labs/community_connector/sources/azure_devops/README.md
@@ -22,12 +22,22 @@ This documentation describes how to configure and use the **Azure DevOps** Lakef
 
 Provide the following **connection-level** options when configuring the connector. These correspond to the connection options exposed by the connector.
 
-| Name                     | Type   | Required | Description                                                                                 | Example                            |
-|--------------------------|--------|----------|---------------------------------------------------------------------------------------------|------------------------------------|
-| `organization`           | string | yes      | Azure DevOps organization name. This is the organization segment in the Azure DevOps URL.  | `my-organization`                  |
-| `project`                | string | no       | Project name or ID. If provided, scopes Git and Work Item operations to this project. If omitted, tables can fetch from all projects. | `my-project`                       |
-| `personal_access_token`  | string | yes      | Personal Access Token (PAT) used for authentication with Azure DevOps Services REST API.   | `7tq...qDnpdg1Nitj8JQQJ99BLAC...` |
-| `externalOptionsAllowList` | string | yes    | Comma-separated list of allowed table-specific options. Must be set to: `project,repository_id,pullrequest_id,status_filter,filter,ids` | `project,repository_id,pullrequest_id,status_filter,filter,ids` |
+| Name | Type | Required | Description | Example |
+|------|------|----------|-------------|---------|
+| `organization` | string | yes (shared) | Azure DevOps organization name — the org segment of the Azure DevOps URL. | `my-organization` |
+| `project` | string | no (shared) | Project name or ID. Scopes Git and Work Item operations to this project; if omitted, tables fetch from all projects. | `my-project` |
+| `externalOptionsAllowList` | string | yes | Comma-separated allowed table-specific options: `project,repository_id,pullrequest_id,status_filter,filter,ids` | `project,repository_id,…` |
+
+**Authentication:** the connector declares two `auth_methods` in its spec — pick **one** when creating the connection:
+
+| Method | Parameters | How it works |
+|--------|-----------|--------------|
+| `pat` | `personal_access_token` (secret) | HTTP Basic auth with a Personal Access Token. Simplest for a single user. |
+| `service_principal` | `tenant_id`, `client_id`, `client_secret` (secret) | Microsoft Entra ID **OAuth 2.0 client-credentials (m2m)** — no interactive user. The Unity Catalog COMMUNITY connection runs the token exchange and refresh **server-side** and injects the bearer token into the connector at query time; the connector never holds the client secret or runs the flow itself. Fits enterprise identity governance and avoids per-user PATs. |
+
+The `service_principal` method uses the spec's `oauth: flow: m2m` block: the token endpoint is `https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token` (the `{tenant_id}` is filled from your `tenant_id` option) and the scope is the Azure DevOps resource `499b84ac-1321-427f-aa17-267ca6975798/.default`.
+
+> **Note:** the `service_principal` (m2m) method is **not yet supported through the community-connector UI** ("Add data" flow). Create the connection with the `community-connector` CLI tool for now (`community-connector create_connection azure_devops <name> -o '{"organization":"…","tenant_id":"…","client_id":"…","client_secret":"…"}'`). The `pat` method works in both the UI and the CLI.
 
 **Important**: The `externalOptionsAllowList` parameter is **required** because tables support table-specific configuration options (`project`, `repository_id`, `pullrequest_id`, `status_filter`, `filter`, `ids`).
 
@@ -53,6 +63,12 @@ Provide the following **connection-level** options when configuring the connecto
        - **Work Items (read)** under the Work Items section - Required if you plan to ingest work item tables
   5. Click **Create** and copy the generated token immediately. Store it securely as you won't be able to see it again.
   6. Use this token as the `personal_access_token` connection option.
+
+- **Entra ID Service Principal (the `service_principal` auth method)**:
+  1. In the Azure portal, register an application (Entra ID → App registrations) — note its **Application (client) ID** and **Directory (tenant) ID**.
+  2. Under **Certificates & secrets**, create a **client secret** and copy its value immediately.
+  3. Add the service principal to your Azure DevOps organization (Organization settings → Users) and grant it the same read access the PAT scopes describe (Code, Graph, Project and Team, Work Items).
+  4. Create the connection with the `service_principal` auth method and supply `tenant_id`, `client_id`, `client_secret` (plus the shared `organization`). Unity Catalog runs the OAuth 2.0 client-credentials exchange and refresh server-side and injects the bearer token at query time — the connector never sees the secret or runs the flow. No PAT required.
 
 ### Create a Unity Catalog Connection
 
--- a/src/databricks/labs/community_connector/sources/azure_devops/connector_spec.yaml
+++ b/src/databricks/labs/community_connector/sources/azure_devops/connector_spec.yaml
@@ -1,13 +1,66 @@
 # azure_devops Community Connector Spec
 # This file defines the specification for the connector including
 # connection parameters and allowlist options.
+#
+# The Azure DevOps connector supports two authentication methods:
+# 1. pat               - Personal Access Token (HTTP Basic).
+# 2. service_principal - Microsoft Entra ID OAuth 2.0 client-credentials
+#                        (m2m). The UC COMMUNITY connection runs the flow and
+#                        injects the bearer token at query time; the connector
+#                        never holds the client secret or runs the exchange.
 
-# ============================================================================
-# Connection Parameters
-# These are provided when creating the Unity Catalog connection.
-# ============================================================================
 display_name: Azure DevOps
 connection:
+  # ---------------------------------------------------------------------------
+  # Authentication Methods
+  # Choose ONE of the following authentication approaches.
+  # ---------------------------------------------------------------------------
+  auth_methods:
+    - name: pat
+      description: >
+        Authenticate with a Personal Access Token via HTTP Basic auth.
+        Simplest for a single user / quick setup.
+      parameters:
+        - name: personal_access_token
+          type: string
+          required: true
+          secret: true
+          description: >
+            Personal Access Token for the Azure DevOps Services REST API.
+            Required scopes: Code (read), Graph (read), Project and Team
+            (read), Work Items (read).
+
+    - name: service_principal
+      description: >
+        Authenticate as a Microsoft Entra ID service principal using the
+        OAuth 2.0 client-credentials (m2m) flow — no interactive user. The
+        service principal must be added to the Azure DevOps organization with
+        the required read access. The token is minted and refreshed by the
+        Unity Catalog connection; the connector receives it as access_token.
+      parameters:
+        - name: tenant_id
+          type: string
+          required: true
+          description: >
+            Microsoft Entra ID tenant (directory) ID. Selects the per-tenant
+            token endpoint used for the client-credentials exchange.
+
+        - name: client_id
+          type: string
+          required: true
+          description: >
+            Entra ID service principal (application) client ID.
+
+        - name: client_secret
+          type: string
+          required: true
+          secret: true
+          description: >
+            Client secret for the Entra ID service principal.
+
+  # ---------------------------------------------------------------------------
+  # Common Parameters (shared across all authentication methods)
+  # ---------------------------------------------------------------------------
   parameters:
     - name: organization
       type: string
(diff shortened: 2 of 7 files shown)


src/databricks/labs/community_connector/sources/azure_devops/connector_spec.yaml
yyoli-db, line 1: if we support multiple auth modes, please do things like 
src/databricks/labs/community_connector/sources/dicomweb/connector_spec.yaml
src/databricks/labs/community_connector/sources/mixpanel/connector_spec.yaml

using `auth_methods`
also, if you use Claude/Cursor or any coding agents with the skill provided in this repo and it did not do the correct thing, please report and we should really tweak the skill to be more effective.

yyoli-db (changes requested):

yyoli-db:
Another thing is that the `COMMUNITY` UC connection supports oauth and several subtypes, like u2m, m2m etc.
Here I feel it is a standard u2m case. Please take a look at the gmail connector and do similar things.

I saw this the oauth endpoints may require tenant_id which is a param user needs to provide, so it is likely we need to enhance the UI a little bit to support this but we could start with the CLI tool.

src/databricks/labs/community_connector/sources/azure_devops/azure_devops_utils.py
```diff
@@ -8,12 +8,72 @@
...
+        self._client_secret = client_secret
+        self._token = None
+        self._expires_at = 0.0
+
+    def _fetch_token(self) -> None:
+        url = (
+            f"https://login.microsoftonline.com/{self._tenant_id}"
+            "/oauth2/v2.0/token"
+        )
+        resp = requests.post(
```
yyoli-db: Minor: token acquisition has no retry. Data-plane calls go through `request_with_retry` (retries 429/500/502/503/504 with backoff), but this token POST does not — a single transient 429/503 from `login.microsoftonline.com` raises `RuntimeError` and aborts the run, including a mid-run refresh during a long `Trigger.AvailableNow` read. Non-blocking (the Entra endpoint is highly available and SDP retries the run), but wrapping this in the same `RETRIABLE_STATUS_CODES` + backoff would make OAuth mode as robust as the data path.

src/databricks/labs/community_connector/sources/azure_devops/azure_devops_utils.py
```diff
@@ -8,12 +8,72 @@
...
+        if resp.status_code != 200:
+            raise RuntimeError(
+                "Azure DevOps OAuth token request failed: "
+                f"{resp.status_code} {resp.text}"
+            )
+        payload = resp.json()
+        self._token = payload["access_token"]
+        self._expires_at = time.time() + int(payload.get("expires_in", 3600))
+
+    def __call__(
```
yyoli-db: Minor (awareness only): the token cache (`self._token` / `self._expires_at`) is read/written here without synchronization. Low risk in practice — this connector declares no partition mixins (the partition test is skipped) and each executor task unpickles its own instance, so reads are effectively single-threaded per reader; worst case under hypothetical concurrency is a redundant token fetch, not wrong auth. A lock is likely overkill unless this connector later gains partitioned/parallel reads.

src/databricks/labs/community_connector/sources/azure_devops/azure_devops_utils.py
```diff
@@ -8,12 +8,72 @@
...
+            data={
+                "grant_type": "client_credentials",
+                "client_id": self._client_id,
+                "client_secret": self._client_secret,
+                "scope": f"{AZURE_DEVOPS_RESOURCE_ID}/.default",
+            },
+            timeout=30,
+        )
+        if resp.status_code != 200:
+            raise RuntimeError(
```
yyoli-db: Trivial: `resp.text` is embedded into the raised `RuntimeError`, so the raw token-endpoint response body reaches logs/exceptions. Fine as-is (Entra AADSTS error bodies don't echo the client secret) and useful for debugging — just flagging it. No change needed.

tests/unit/sources/azure_devops/test_azure_devops_lakeflow_connect.py
```diff
@@ -12,3 +23,102 @@ class TestAzureDevopsConnector(LakeflowConnectTests):
...
+            return _FakeTokenResponse(token="tok-1")
+
+        monkeypatch.setattr(_POST_PATH, fake_post)
+        req = EntraClientSecretAuth("tenant", "client", "secret")(self._prepared())
+        assert req.headers["Authorization"] == "Bearer tok-1"
+        assert "login.microsoftonline.com/tenant/oauth2/v2.0/token" in calls[0][0]
+        assert calls[0][1]["grant_type"] == "client_credentials"
+        assert calls[0][1]["scope"].endswith("/.default")
+
+    def test_reuses_token_until_expiry(self, monkeypatch):
```
yyoli-db: Minor (test gap): this asserts the token is fetched once and reused while valid, but nothing forces `_expires_at` into the past (or within the 300s skew) to prove a second fetch occurs. The refresh-before-expiry path — the whole reason OAuth mode survives the ~1h TTL — is currently only verified by inspection. Suggest adding a case that pushes `_expires_at` to `time.time()` (or monkeypatches `time.time` / sets a tiny `expires_in`) and asserts the fetch count goes 1 -> 2 and the new bearer is applied.

wesley84:
# Self-Review — azure_devops

**Overall: 89 / 100 — ALMOST**

Run at: 2026-07-09 · Scope context: PR #232 (service-principal m2m OAuth rework)

> **PR #232 verdict:** the auth rework is **clean** — `auth_methods` (pat +
> service_principal), OAuth offloaded to UC (`flow: m2m`), the connector reads
> the UC-injected `access_token`; pylint 10/10; validated live on e2-dogfood.
> Every finding below is **pre-existing connector behavior**, not introduced by
> #232. No BLOCKER fails.

## Top recommendations (all pre-existing)

1. **MAJOR — A5** No `_init_time` upper-bound cap on `append` tables. They rely
   on server-side cursor filtering; `test_read_terminates` passes, so
   termination holds today, but a long `AvailableNow` trigger could chase
   continuously-arriving rows. *Fix: freeze an init-time cursor bound.*
2. **MAJOR — A6** `max_records_per_batch` not consumed (`azure_devops.py`) —
   `append` tables paginate server-side without a client admission cap. Fine for
   small orgs; can be heavy on large repos. *Fix: honor `max_records_per_batch`.*
3. **MAJOR — B11** No record-mode validation artifact on file (gitignored;
   expected on a fresh checkout). *Fix: `CONNECTOR_TEST_MODE=record … pytest`.*

## A. Connector implementation — 21 / 28
- ✅ A1. Extends `LakeflowConnect` — `azure_devops.py:28`
- ✅ A2. Abstract methods present — `list_tables:95, get_table_schema:98, read_table_metadata:105, read_table:112`
- ✅ A3. No `cdc_with_deletes` tables → `read_table_deletes` not required (N/A)
- ✅ A4. Read pattern: `append` (cursor) + `snapshot` — deliberate
- ⚠️ A5. No `_init_time` cap on append tables — see Top recs (pre-existing)
- ⚠️ A6. `max_records_per_batch` not honored — see Top recs (pre-existing)
- ✅ A7. All HTTP calls have `timeout=` (e.g. `azure_devops.py:918-922`)
- ⚠️ A8. `IntegerType` used for changeCounts Add/Edit/Delete — `azure_devops_schemas.py:80-82` (minor; small ints)
- ✅ A9. No nested-field flattening
- ✅ A10. Imports clean (stdlib / requests / pyspark / interface only)
- ✅ A11. Pylint 10.00/10 on source files (the `__init__.py` F0001 is a local namespace-package tooling artifact, not a code finding)
- ✅ A12. No ingestion-agent surface implemented
- ✅ A13. `AzureDevopsDataSource(LakeflowSource)` + `_lakeflow_connect_cls` — `__init__.py:11-12`

- ✅ B1–B7. Test file, `LakeflowConnectTests`, `connector_class`, `simulator_source`, `replay_config` (org/project/PAT — matches the pat auth path), `endpoints.yaml`, 15 corpus files
- ✅ B8. Connector URLs resolve against spec (simulate suite passes)
- ✅ B9. `test_read_terminates` passes (simulate)
- ✅ B10. Full suite: 33 passed, 2 skipped (simulate)
- ❌ B11. No record-mode artifacts on file — see Top recs (pre-existing)
- ⚠️ B12. Coverage not assessable (no `coverage.json` — tied to B11)
- ➖ B13. `synthesize_future_records` N/A — no `_init_time` cap to exercise (see A5)
- ✅ B14. Read-only connector (no write-back) → `test_utils_class` correctly unset

## C. Artifacts — 28 / 28
- ✅ C1. `azure_devops.py` compiles
- ✅ C2. `azure_devops_api_doc.md` present
- ✅ C3. `connector_spec.yaml` parses; `auth_methods` + shared params; `oauth` block valid (`flow: m2m`; `client_id`/`client_secret`/`tenant_id` listed; UC-injected `access_token` correctly NOT a spec param)
- ✅ C4. `README.md` present
- ✅ C5. `pyproject.toml` — `requests` in runtime deps
- ✅ C6. `_generated_azure_devops_python_source.py` regenerated, fresh
- ✅ C8. 15 corpus files valid JSON
- ✅ C9. Test class present
- ✅ C10. `pyspark` in `[optional-dependencies].dev` only — not a runtime dep

## D. Security — 21 / 22
- ✅ D1. No hardcoded secrets (connector no longer holds the client secret — UC does)
- ✅ D2–D9. No eval/exec, subprocess, `verify=False`, unsafe `yaml.load`, `pickle.load`, path traversal, secrets-in-logs, or `http://`
- ⚠️ D10. `lakeflow-community-connectors>=0.1.0` is unbounded (no upper pin) — `pyproject.toml`

- ✅ E1. `list_tables` ≡ API doc ≡ README (spot-checked)
- ✅ E2. Connection params consistent (OAuth exemption applied: UC-injected `access_token` absent from spec/README as expected; `client_id`/`client_secret`/`tenant_id`/`organization`/`project`/`personal_access_token` agree)
- ✅ E3. Schemas subset of API doc (spot-check)
- ✅ E4. Primary keys consistent
- ✅ E5. `external_options_allowlist` (project, repository_id, pullrequest_id, status_filter, filter, ids) ≡ `table_options.get(...)` keys
- ✅ E7. No README/API-doc contradictions found

src/databricks/labs/community_connector/sources/azure_devops/connector_spec.yaml
wesley84: Addressed all of this — thanks for the steer.

Reworked to standard `auth_methods` (pat + service_principal) per dicomweb/mixpanel, and **offloaded the OAuth flow to UC** instead of hand-rolling it.

One clarification on **u2m vs m2m**: the driver (Rockwell) needs a service principal + client secret with no interactive user, so this is the **m2m** client-credentials subtype rather than u2m/gmail. **Validated live on e2-dogfood** 🎉 — UC minted the m2m token from the SP and injected it, and the connector read the ADO `projects` table (3 rows). First m2m path exercised end-to-end.

The framework gap you predicted was real: the spec couldn't template the per-tenant token endpoint, so I added `{tenant_id}` interpolation in the CLI's oauth-option resolution (+ a test).

PR is rebased on current master; self-review posted above (ALMOST/89, no blockers — flagged items are pre-existing). Could you re-review and apply `safe-to-test` / `connector-self-reviewed` when you get a chance? My fork account can't set the labels.

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/azure_devops/README.md
```diff
@@ -24,10 +24,20 @@ Provide the following **connection-level** options when configuring the connecto
 
 | Name                     | Type   | Required | Description                                                                                 | Example                            |
```
yyoli-db: This seems some orphaned header, please clean up.

src/databricks/labs/community_connector/sources/azure_devops/connector_spec.yaml
```diff
@@ -24,14 +77,21 @@ connection:
...
+  #                    server-side and injects `access_token` at query time.
+  #   token_url      — per-tenant token endpoint. `{tenant_id}` is interpolated
+  #                    from the user-supplied tenant_id option by the CLI
+  #                    (see the connector-spec parameter-templating support).
+  #   scopes         — the Azure DevOps resource `.default` scope; the fixed
+  #                    GUID is the first-party Azure DevOps application ID.
+  # ---------------------------------------------------------------------------
+  oauth:
+    flow: m2m
+    token_url: https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token
```
yyoli-db, line 93: this won't be immediately supported through the UI. so please update the README doc to have a note. Using CLI tool is fine but if create connection using community connector UI, this won't work for now.
@rahulkhanna-dbl

yyoli-db (changes requested):
Looks good. Please address the 2 comments and we will merge.

src/databricks/labs/community_connector/sources/azure_devops/README.md
wesley84: Fixed in 5410354 — removed the duplicated/orphaned table header.

src/databricks/labs/community_connector/sources/azure_devops/connector_spec.yaml
wesley84: Added a README note in 5410354: the service_principal (m2m) method isn't supported via the community-connector UI yet — create the connection with the CLI `create_connection` for now; the PAT method works in both the UI and CLI.

yyoli-db (approved):

PR #240, opened 2026-07-08
fix(azure-devops): workitems read returns empty batch when no watermark change

sansbacon:
The pipeline would fail immediately on every trigger after initial load,
making it permanently broken in steady state.

## Root Cause

`_read_workitems` fetches all current work items on every call (WIQL
`ChangedDate` filtering is date-only and rejects timestamp literals, so
fine-grained server-side filtering isn't available). The watermark is
the max `System.ChangedDate` across returned items, and advances only
when an item changes.

The bug: when no watermark advanced (no items changed), the method still
returned the full work-item list as a non-empty batch. The `SimpleDataSourceStreamReader`
framework requires that if `read()` returns a non-empty batch, the end
offset must strictly advance past the start offset.

## Fix

Track a `watermarks_advanced` flag across the project loop. When no
project watermark advances, return `iter([]), start_offset` (empty batch,
same offset) — the correct signal to the streaming framework that there
is no new data. The full batch is only returned when at least one
watermark advances.

Applied to both `azure_devops.py` (source) and
`_generated_azure_devops_python_source.py` (generated file).

## Testing

Verified on a live Azure DevOps organization (`DIA_Products` project,
1000+ work items). The pipeline completed cleanly after previously
failing on every trigger for 2+ days (19 consecutive failures). No
issues reported post-fix. A full refresh completed in ~2 minutes with
no errors.


yyoli-db:
@sansbacon some unit tests would be valuable here. But I will let the PR gets merged first as the fix seems obvious and no harm. Please do a follow up PR for the test

yyoli-db (approved):

PR #249, opened 2026-07-21
feat(mailchimp): add Mailchimp Marketing API connector

kawaiiblitz:
## Summary
Adds a Lakeflow Community Connector for the Mailchimp Marketing API.

Tables (all CDC): campaigns, reports, lists, members.

- Auth: API key (HTTP Basic), datacenter derived from the key suffix.
- Incremental reads: sliding window capped at _init_ts, with a
  configurable lookback_seconds to include the exclusive since_*
  boundary row. members is fanned out lazily per list_id.

## Testing
- Simulate-mode: 7 passed / 2 skipped.
- Live/record-mode validated end-to-end against a real Mailchimp
  account: 4/4 endpoints validated, 0 mismatches.

## Notes
- Ready for safe-to-test labeling.


kawaiiblitz:
# Self-Review — mailchimp

**Overall: 100 / 100 — READY** — no BLOCKER failures.

Run at: 2026-07-23T23:40:08Z

Re-run after addressing the two new review findings on PR #249:
same-second cursor cluster **drain-on-stall**, and **full pagination** of the
unsorted oldest-cursor probe. Static audit plus pylint and the full pytest
suite executed on serverless (env client "3"). No BLOCKER or MAJOR findings.

## Top recommendations

None blocking. Ready to ship pending the maintainer re-applying
`safe-to-test` for CI.

## A. Connector implementation — 33 / 33
- ✅ A1. Class extends LakeflowConnect — `mailchimp.py`
- ✅ A2. All abstract methods implemented (`list_tables`, `get_table_schema`, `read_table_metadata`, `read_table`)
- ✅ A3. No `cdc_with_deletes` table → `read_table_deletes` not required
- ✅ A4. Read pattern: sliding window (Strategy A); nested fan-out for `members`
- ✅ A5. `_init_ts` cap present with timestamp-aware guard — `mailchimp.py:358`
- ✅ A6. `max_records_per_batch` honored on sorted endpoints; unsorted endpoints drain the window; same-second cluster re-drains uncapped so no row is lost — `mailchimp.py:384-407`
- ✅ A7. Single HTTP call goes through `_request_with_retry` with `timeout=60` — `mailchimp.py:120`
- ✅ A8–A10. Schemas use LongType/StructType; nested fields not flattened; imports clean
- ✅ A11. Pylint **10.00/10** (CI flag set, run on serverless) — RC=0
- ✅ A12. No experimental ingestion-agent surface
- ✅ A13. `MailchimpDataSource(LakeflowSource)` exposed with `_lakeflow_connect_cls` — `__init__.py`

## B. Testing & simulator validation — 26 / 26
- ✅ B1–B8. Test file, subclass, config, spec, corpus all present; every URL has a spec entry
- ✅ B9–B10. **15 passed / 2 skipped** in simulate mode on serverless (RC=0), incl. the 2 new tests this round (same-second cluster; unsorted-probe full pagination)
- ✅ B11. Live/record run on file, status clean (0 drift), 4/4 endpoints validated
- ✅ B12. Coverage full
- ✅ B13. `synthesize_future_records` present on cursor-bearing endpoints
- ✅ B14. No write-back → `test_utils_class` not required

## C. Artifacts — 20 / 20
- ✅ C1. py_compile clean
- ✅ C2–C5. API doc, spec, README, pyproject all present and consistent
- ✅ C6. `_generated_mailchimp_python_source.py` regenerated (newer than source)
- ✅ C7–C8. Simulator spec + corpus valid
- ✅ C9. Test class present
- ✅ C10. No runtime `pyspark` (dev-only optional dependency)

## D. Security — 15 / 15
- ✅ D1. No hardcoded secrets
- ✅ D2–D3. No eval/exec/subprocess
- ✅ D4. No `verify=False`
- ✅ D5–D9. No pickle, safe_load YAML, no secret logging, no `http://` base URLs
- ✅ D10. Dependencies pinned (`requests>=2.32.0`)

## E. Cross-doc consistency — 9 / 9
- ✅ E1. Tables agree across code / API doc / README
- ✅ E2. Connection params (`api_key`, `server_prefix`) agree across all three
- ✅ E3–E4. Schemas and primary keys match API doc
- ✅ E5. `external_options_allowlist` = `max_records_per_batch,window_seconds,start_timestamp,lookback_seconds` — matches all four `table_options.get(...)` keys read by the code
- ✅ E6–E7. README quick-start and semantics consistent

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
```diff
@@ -0,0 +1,458 @@
...
+                    continue
+                cursor = self._peek_from_path(
+                    cfg["path"].format(list_id=list_id),
+                    cfg["resource_key"],
+                    cursor_field,
+                    supports_sort=cfg["supports_sort"],
+                )
+                if cursor:
+                    return cursor
+            return None
```
yyoli-db: **`members` auto-discovery starts from the first audience's oldest member, not the global oldest.** This branch returns the first non-null cursor found (`return cursor`) rather than the min across audiences. Since all members share one cursor window, seeding `since` from audience A's oldest `last_changed` can permanently skip any member in audience B with an older `last_changed` (the exclusive `since_*` filter excludes it on every subsequent window). This contradicts the README's "backfills all history" claim. Consider taking the min cursor across all probed audiences, or documenting that `start_timestamp` is effectively required for full `members` backfill.

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
```diff
@@ -0,0 +1,458 @@
...
+        )
+
+        # Drained the whole window (drained fully unless we bailed on the cap).
+        window_drained = len(records) < max_records
+        if window_drained:
+            end_offset = {"cursor": window_end}
+        else:
+            # Cap hit mid-window: resume from the max cursor seen. CDC upsert
+            # tolerates the re-fetch of the window head on the next call.
+            end_offset = {"cursor": self._max_cursor(records, cursor_field, window_end)}
```
yyoli-db, line 417: **Non-sorted endpoints can skip rows on a mid-window cap.** Resuming `end_offset` from `_max_cursor(records, ...)` is only safe when records were fetched in ascending cursor order (`supports_sort=True`: campaigns, members). `reports` and `lists` set `supports_sort=False`, so the fetched subset is in server-default order — resuming from the max cursor seen can drop unfetched rows whose cursor is below that max. Triggering it needs >1000 rows in one window for a low-cardinality table, so the risk is small, but it is real data loss when it fires. Consider using the `_max_cursor` resume path only for sorted endpoints.

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
```diff
@@ -0,0 +1,458 @@
...
+        if cfg["supports_sort"]:
+            base_params["sort_field"] = cursor_field
+            base_params["sort_dir"] = "ASC"
+
+        records = self._paginate(
+            cfg["path"], cfg["resource_key"], base_params, max_records
+        )
+
+        # Drained the whole window (drained fully unless we bailed on the cap).
+        window_drained = len(records) < max_records
```
yyoli-db: **`window_drained` misfires when a window holds between `max_records` and `count` rows.** With `count=1000` and `max_records=200`, a 300-row window is fully fetched in one short page, yet `300 < 200` is False, so it resumes from `max_cursor` instead of advancing to `window_end`. No data loss for sorted endpoints, but it causes redundant re-fetching and keeps the cursor from cleanly reaching `window_end`. The "drained" signal really wants to compare against whether the last page was short, not against `max_records`.

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
```diff
@@ -0,0 +1,458 @@
...
+        cursor_field = cfg["cursor_field"]
+
+        since = self._resolve_since(table_name, start_offset, table_options, cfg)
+        if not since:
+            raise ValueError(
+                f"Cannot determine a starting cursor for '{table_name}'. "
+                f"Provide 'start_timestamp' in table_options."
+            )
+        # Already caught up to init time — nothing new this run.
+        if since >= self._init_ts:
```
yyoli-db: **String comparison mixes `Z` and `+00:00` offsets.** `since >= self._init_ts` (and `min(end, self._init_ts)` in `_window_end`) compare ISO timestamps as strings. `_init_ts` is always `...+00:00`.

It is much safer to parse both sides to `datetime`, or normalize `since` once at resolve time.

src/databricks/labs/community_connector/sources/mailchimp/README.md
```diff
@@ -0,0 +1,267 @@
...
+|-------------|--------|----------|------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------|
+| `api_key`   | string | yes      | Mailchimp Marketing API key used for HTTP Basic authentication. The account data center is derived from the key's `-<dc>` suffix.  | `0123456789abcdef0123456789abcde-us21`   |
+| `server_prefix` | string | no   | Data center / server prefix (e.g. `us21`). Only needed for **legacy** keys that have no `-<dc>` suffix; otherwise it is derived automatically from `api_key`. | `us21`                                   |
+| `externalOptionsAllowList` | string | yes | Comma-separated list of table-specific option names that are allowed to be passed through to the connector. This connector supports table-specific options, so this parameter must be set. | `max_records_per_batch,window_seconds,start_timestamp` |
+
+The full list of supported table-specific options for `externalOptionsAllowList` is:
+`max_records_per_batch,window_seconds,start_timestamp`
```
yyoli-db: **[Low] README omits `lookback_seconds` from the documented allowlist.** `connector_spec.yaml` declares `...,lookback_seconds` and the option has real behavior, but the README allowlist strings (here and lines 26, 58) stop at `start_timestamp`. A user copying this string verbatim can't pass `lookback_seconds`. Please add it here and to the Source-specific `table_configuration` options table.

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
```diff
@@ -0,0 +1,458 @@
...
+
+        429 is Mailchimp's rate-limit signal (max 10 concurrent connections
+        per key); 500/503 are transient server errors.
+        """
+        url = f"{self._base_url}{path}"
+        backoff = INITIAL_BACKOFF
+        resp = None
+        for attempt in range(MAX_RETRIES):
+            resp = requests.get(
+                url, params=params or {}, auth=self._auth(), timeout=60
```
yyoli-db: **Transport errors aren't retried.** `_request_with_retry` retries only on HTTP status codes; a `ConnectionError`/`Timeout` propagates immediately. Since the backoff loop already exists, wrapping `requests.get` to retry transient transport errors would make executor-side reads more resilient. Optional.

yyoli-db (changes requested):

kawaiiblitz:
Thanks for the thorough review @yyoli-db! Addressed all six points in the latest push. Summary:

1. **`members` global oldest cursor** — `_peek_oldest_cursor` now probes every audience and takes the minimum `last_changed` across all of them, instead of returning the first audience's oldest. This prevents permanently skipping older members in other audiences under the exclusive `since_*` filter.

2. **Unsorted endpoints skipping rows on a mid-window cap** — `reports`/`lists` (`supports_sort=False`) now drain the full window rather than truncating at `max_records` and resuming from the max cursor seen. `_paginate` returns an `exhausted` flag so the caller only takes the `_max_cursor` resume path on sorted endpoints. These tables are low-cardinality and the window bounds the volume.

3. **`window_drained` misfiring** — the drained signal is now the `exhausted` flag from `_paginate` (short page / total reached), not `len(records) < max_records`. A sorted window between `max_records` and `count` now advances cleanly to `window_end`.

4. **`Z` vs `+00:00` string comparison** — added a shared `_to_dt` helper; all cursor comparisons (`since >= _init_ts`) and arithmetic (`_window_end`, `_apply_lookback`) now parse to timezone-aware datetimes.

5. **README omits `lookback_seconds`** — added to the allowlist strings (3 spots) and to the `table_configuration` options table.

6. **Transport errors not retried** — `_request_with_retry` now retries `ConnectionError`/`Timeout` on the same backoff and re-raises after the last attempt.

Added `tests/unit/sources/mailchimp/test_mailchimp_unit.py` with regression tests reproducing each scenario (global-oldest probe, unsorted-window drain, drained-flag, `Z`/`+00:00` compare). Simulate suite: **13 passed / 2 skipped** on serverless. Ready for another look when you have a chance.

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
```diff
@@ -0,0 +1,502 @@
...
+        )
+
+        if window_drained:
+            # Endpoint fully drained for this window: advance cleanly to the end.
+            end_offset = {"cursor": window_end}
+        else:
+            # Cap hit mid-window on a sorted endpoint: resume from the max cursor
+            # seen (everything below it is already fetched). CDC upsert tolerates
+            # the re-fetch of the window head on the next call.
+            end_offset = {"cursor": self._max_cursor(records, cursor_field, window_end)}
```
yyoli-db, line 417: same-second cursor cluster larger than a page can stall and drop rows.

When the max_records cap is hit mid-window and more records than one page share an identical cursor timestamp, _max_cursor returns a value equal to the window's since. The next call re-issues the same query, re-fetches the same first page, and produces end_offset == start_offset, so pagination halts and everything past the first page is lost.

Reproduced locally: 1500 campaigns all at create_time=2026-03-01T00:00:00+00:00 -> 1000 ingested, 500 dropped permanently. Unlikely for campaigns but realistic for members (bulk import stamps many contacts with the same last_changed second). Consider detecting max_cursor == since after a cap hit and draining the window instead of resuming, or documenting the 'one cursor-second must fit in one page' constraint.

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
```diff
@@ -0,0 +1,502 @@
...
+                "offset": 0,
+                "sort_field": cursor_field,
+                "sort_dir": "ASC",
+            }
+            batch, _ = self._fetch_page(path, resource_key, params)
+            if batch:
+                return batch[0].get(cursor_field)
+            return None
+        # No sort support: take the min cursor across the first full page.
+        batch, _ = self._fetch_page(
```
yyoli-db: oldest-cursor probe on unsorted endpoints only sees the first page.

For reports/lists (supports_sort=False), this fetches a single page (count=1000, offset=0) and takes the min cursor across just that page. If the account has >1000 reports and the true-oldest row sits on a later page in server-default order, since is seeded too late and older rows are excluded by the exclusive since_* filter on every subsequent window.

Reproduced locally: 1200 reports with the global-oldest at index 1100 -> probe returns the page-1 min and the older row is never ingestible. Setting start_timestamp sidesteps it; consider paginating the probe fully for unsorted endpoints, or at least a comment noting the first-page assumption.

yyoli-db (changes requested):

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
yyoli-db: **Suggested fix — drain-on-stall (mirrors the unsorted-endpoint philosophy).** When the cap trips but the resume cursor can't advance past the window start, re-drain the whole window uncapped. `window_seconds` bounds the volume, so peak memory stays bounded and forward progress is guaranteed:

```python
records, window_drained = self._paginate(
    cfg["path"], cfg["resource_key"], base_params, page_cap
)

# Same-second cluster larger than the cap: the resume cursor would equal
# `since`, so an offset/time-cursor resume can't make progress. Re-drain the
# whole window uncapped — window_seconds bounds the volume. (Only reachable on
# sorted endpoints; the unsorted path never sets page_cap.)
if not window_drained and page_cap is not None:
    resume = self._max_cursor(records, cursor_field, window_end)
    if self._to_dt(resume) <= self._to_dt(since):
        records, window_drained = self._paginate(
            cfg["path"], cfg["resource_key"], base_params, None
        )
```

Alternatively (lighter weight), document the constraint that a single cursor-second must fit within one `max_records_per_batch` page — but that is a silent-data-loss trap for `members` bulk imports, so the code fix is preferred. Worth a unit test: 1500 records all at one `create_time` with `max_records_per_batch < 1500`, assert all 1500 ingest across successive `_read_window` calls.

src/databricks/labs/community_connector/sources/mailchimp/mailchimp.py
yyoli-db: **Suggested fix — fully paginate the unsorted probe.** `_paginate` already drains all pages when `max_records is None`, and this probe only runs on first-run discovery, so it's a one-time cost:

```python
# No sort support: drain all pages and take the global min cursor. (This runs
# only on first-run discovery, so the full enumeration is a one-time cost.)
batch, _ = self._paginate(
    path, resource_key, {"count": DEFAULT_PAGE_COUNT}
)
cursors = [r.get(cursor_field) for r in batch if r.get(cursor_field)]
return min(cursors) if cursors else None
```

This also removes the "small in cardinality" assumption in the docstring. Worth a unit test: 1200 `reports` with the global-oldest at index 1100, assert `_peek_oldest_cursor` returns that oldest cursor.

yyoli-db (approved):

PR #254, opened 2026-07-25
feat(collibra): add Collibra community connector

wesley84:
## What's included
- API research doc: `sources/collibra/collibra_api_doc.md`
- Implementation: `sources/collibra/collibra.py` (+ `collibra_schemas.py`, `collibra_utils.py`, `__init__.py`, `pyproject.toml`)
- Simulator spec + corpus: `source_simulator/specs/collibra/`
- Connector spec: `sources/collibra/connector_spec.yaml`
- Docs: `sources/collibra/README.md`
- Tests: `tests/unit/sources/collibra/`

Read-only Collibra Data Intelligence Platform connector — extracts governed metadata (descriptions, owners/stewards, certification, taxonomy) into UC Delta tables. This is the ingress half of a catalog→UC metadata-hydration solution (design validated with the Metadata Intelligence PMs).

## Scope (lean v1) — 4 tables
| table | ingestion | PK | cursor |
|---|---|---|---|
| `assets` | cdc | id | lastModifiedOn |
| `attributes` | cdc | id | lastModifiedOn |
| `responsibilities` | cdc | id | lastModifiedOn |
| `domains` | snapshot | id | — |

`relations` / `communities` / `users` deferred until a consumer needs them.

## Auth
OAuth 2.0 client-credentials (m2m), offloaded to the UC COMMUNITY connection — reuses the pattern from the Azure DevOps connector (#232). Connector consumes a UC-injected `access_token` (Bearer); static token fallback supported.

## Simulate-mode test results
`pytest tests/unit/sources/collibra/` → **7 passed, 2 skipped** (skips: partition mixin N/A, no `cdc_with_deletes`). pylint 10/10. All 4 tables read, paginate, and terminate; CDC offsets converge on 2nd call.

The simulator spec + corpus have **not** yet been validated against the live API.

## Next step: live testing
Awaiting Collibra API access on the ISV demo instance (`databricks.collibra.com`). Once available:
`/validate-connector collibra`

Live-only unknowns flagged `needs-live-testing` in code: Description/Certification `typePublicId`, role UUIDs, `/assets` cursor mechanism.


wesley84:
# Self-Review — collibra

**Overall: 88 / 100 — ALMOST**

Run at: 2026-07-25 · Read-only Collibra → UC metadata connector (4 tables, lean v1)

> No BLOCKER fails. One MAJOR is a **framework-level** m2m token-expiry limitation
> on large full-loads (documented, raised with the connectors team — not a
> per-connector bug). The rest is minor polish.

## Top recommendations

1. **MAJOR (framework, not connector) — m2m token expiry on large full-loads.**
   The UC-injected `access_token` is read once in `__init__` (`collibra.py:115`
   area, `_configure_auth`) and cached in the session header — the standard m2m
   pattern. Over a long `Trigger.AvailableNow` drain of a large table (validated:
   `assets`, 264K rows), the run outlives the ~1h token → `401 expiredToken`.
   `max_records_per_batch` caps batch *size*, not total *runtime*. **Affects any
   m2m connector (ADO included)**; needs a UC/framework token-refresh answer.
   Snapshot + incremental tables unaffected (`domains` validated end-to-end).
2. **MINOR — A8** `ratingsCount` uses `IntegerType` — `collibra_schemas.py:71`.
   Bounded rating count so overflow is implausible; consider `LongType` for
   consistency with the connector's "Long not Integer" convention.
3. **MINOR — B11** no record-mode cassette on file. Validated instead via a real
   e2-dogfood pipeline run (domains, 3,296 rows) — stronger than a cassette, but
   the simulator record artifact isn't present. Optional to add later.

## A. Connector implementation — 27 / 31
- ✅ A1. Extends `LakeflowConnect` — `collibra.py:61`
- ✅ A2. Abstract methods present — `list_tables:146, get_table_schema:149, read_table_metadata:155, read_table:161`
- ✅ A3. No `cdc_with_deletes` tables → `read_table_deletes` not required (N/A)
- ✅ A4. Read pattern: cdc (assets/attributes/responsibilities) + snapshot (domains); standard (non-partitioned) — deliberate, Collibra has no server-side time filter
- ✅ A5. Init-time cap present — `_init_ts` `collibra.py:115`
- ✅ A6. `max_records_per_batch` honored on incremental tables — `collibra.py:427`
- ✅ A7. All HTTP calls have `timeout=` (collibra_utils request helpers)
- ⚠️ A8. `ratingsCount` uses `IntegerType` — `collibra_schemas.py:71` (minor)
- ✅ A9. Schemas use `StructType` refs, not flattened
- ✅ A10. Imports clean (stdlib / requests / pyspark / interface only)
- ✅ A11. Pylint 10.00/10
- ✅ A13. `CollibraDataSource(LakeflowSource)` + `_lakeflow_connect_cls` — `__init__.py:9-10`

- ✅ B1–B7. Test file, `LakeflowConnectTests`, connector_class, `simulator_source="collibra"`, `replay_config` (org + access_token — matches `_configure_auth`), endpoints.yaml, 4 corpus files
- ✅ B8. Connector URLs resolve against spec (simulate passes)
- ✅ B9. `test_read_terminates` passes
- ✅ B10. Full suite: 7 passed, 2 skipped
- ⚠️ B11. No record-mode cassette — validated via live e2-dogfood pipeline instead (domains end-to-end). Minor.
- ✅ B12. Coverage — n/a (tied to B11)
- ✅ B13. `synthesize_future_records` present in endpoints.yaml (exercises the `_init_ts` cap)
- ✅ B14. Read-only connector → `test_utils_class` correctly unset

## C. Artifacts — 27 / 27
- ✅ C1. `collibra.py` compiles
- ✅ C2. `collibra_api_doc.md` present (+ live-validation findings)
- ✅ C3. `connector_spec.yaml` parses; `auth_methods` (token, service_principal) + `oauth: flow: m2m` (no scope — live-validated); UC-injected `access_token` correctly NOT a spec param
- ✅ C4. `README.md` present
- ✅ C5. `pyproject.toml` — requests runtime dep
- ✅ C6. `_generated_collibra_python_source.py` regenerated, fresh
- ✅ C8. 4 corpus files valid JSON
- ✅ C9. Test class present
- ✅ C10. `pyspark` in `[optional-dependencies].dev` only — not runtime

## D. Security — 21 / 22
- ✅ D1–D9. No hardcoded secrets, eval/exec, subprocess, verify=False, unsafe yaml.load, pickle, path traversal, secrets-in-logs, or http://
- ⚠️ D10. `lakeflow-community-connectors>=0.1.0` unbounded (no upper pin) — `pyproject.toml`

- ✅ E1. `list_tables` (assets/attributes/responsibilities/domains) ≡ API doc ≡ README
- ✅ E2. Connection params consistent (OAuth exemption: UC-injected `access_token` absent from spec/README as expected; org/base_url/client_id/client_secret/token agree)
- ✅ E3. Schemas subset of API doc
- ✅ E4. Primary keys (`id`) consistent
- ✅ E5. `external_options_allowlist` ≡ all `table_options.get(...)` keys (sort_field, domain_id, community_id, asset_id, type_public_ids, resource_ids, role_ids, max_records_per_batch, page_size)
- ✅ E7. No README/API-doc contradictions

wesley84:
Retitled — dropped the `[needs-live-testing]` tag. Since opening, this has been **live-validated on e2-dogfood**: UC minted the Collibra m2m token and the `domains` table landed end-to-end (3,296 rows); all 4 endpoints confirmed 200 against the live instance; 2 live-caught bugs fixed (OAuth scope must be omitted, and `/assets` sortField=ID not LAST_MODIFIED). One documented limitation remains — large cdc full-loads (e.g. `assets`, 264K) can outlive the ~1h UC-injected m2m token; that's a framework-level token-refresh question (affects any m2m connector), raised separately. Snapshot + incremental are unaffected.

src/databricks/labs/community_connector/sources/collibra/collibra.py
```diff
@@ -0,0 +1,455 @@
...
+            # Init-time cap: skip records modified after the connector started.
+            if cursor is not None and cursor > self._init_ts:
+                continue
+
+            rec = transform(raw) if transform else self._shape_common(raw)
+            records.append(rec)
+            if cursor is not None and (max_seen is None or cursor > max_seen):
+                max_seen = cursor
+
+            if max_records is not None and len(records) >= max_records:
```
yyoli-db, line 387: `assets` silently drops the bulk of the collection during incremental sync.

`_read_assets` fetches sorted by `sortField=ID` (line 196; `LAST_MODIFIED` is rejected 400 on `/assets`), so records arrive in **ID order**. But `_incremental_from_iter` assumes ascending `lastModifiedOn` (docstring line 307) and, when the loop truncates here at `max_records` (default 5000), advances the offset to `max_seen` = the max `lastModifiedOn` among that ID-ordered subset. On the next call the strict `> since` filter (line 337) permanently skips every not-yet-emitted asset whose `lastModifiedOn <= max_seen`. With ~264k assets and a 5000 cap the first batch cannot cover the collection, so loss is guaranteed and silent.

Fix: for the ID-ordered `assets` path `lastModifiedOn` is not a keyset, so no count-based truncation can safely advance the watermark. Force `max_records=None` for `assets` (rely on the `_init_ts` upper cap for termination), or switch `assets` to a real ID-keyset resume cursor (store last-seen `id`, filter `id > last_id`).

src/databricks/labs/community_connector/sources/collibra/collibra.py
```diff
@@ -0,0 +1,455 @@
...
+                break
+
+        if not records or max_seen is None or max_seen == since_val:
```
yyoli-db, line 390: affects `attributes`/`responsibilities` (which are server-sorted `LAST_MODIFIED ASC`, so the gross loss in the assets case does not apply here).

The loop breaks exactly at `max_records` (lines 348-349). If records past the boundary share the same `lastModifiedOn` as the last emitted record, `max_seen` equals their cursor, and the next batch's strict `> since_val` filter (line 337) permanently skips them. Classic keyset-on-non-unique-key + strict-inequality + mid-tie-group truncation defect.

Fix: treat `max_records` as a soft limit — keep draining while `cursor == max_seen` before breaking, so a batch never splits a group sharing one `lastModifiedOn`.

src/databricks/labs/community_connector/sources/collibra/README.md
```diff
@@ -0,0 +1,301 @@
...
+- **Domain and community descriptions** are native string fields on the domain object (`domains.description`), unlike asset descriptions.
+
+### Required and optional table options
+
+All tables work with no options. Optional table-specific options narrow the extract:
+
+| Table              | Required Options | Optional Options                                          | Notes |
+|--------------------|------------------|-----------------------------------------------------------|-------|
+| `assets`           | None             | `sort_field`, `domain_id`, `community_id`, `max_records_per_batch` | `sort_field` defaults to `LAST_MODIFIED` (needed for the incremental cursor); `domain_id` / `community_id` scope to a single domain or community. |
```
yyoli-db: this says `sort_field` defaults to `LAST_MODIFIED`, but the code defaults to `ID` (collibra.py:196) and `LAST_MODIFIED` is rejected 400 on `/assets` (api_doc gotcha #2). A user following this and setting `sort_field=LAST_MODIFIED` would get a 400. Change to `ID`.

src/databricks/labs/community_connector/sources/collibra/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Step 3: Run the pipeline
+
+The first run does a full backfill across all tables (the CDC tables fetch all historical records, since they have no checkpoint to resume from). Subsequent runs ingest only records with `lastModifiedOn` greater than each table's saved watermark. `domains` is re-snapshotted in full each run.
+
+## Incremental sync behavior
+
+Because Collibra's Core REST API has **no server-side `lastModifiedAfter` filter**, the connector implements incremental sync client-side:
+
+- The `cdc` readers request pages sorted by `LAST_MODIFIED ASC` and apply the saved cursor as a strict `> since` filter, so the boundary record is not re-emitted on resume.
```
yyoli-db: "The `cdc` readers request pages sorted by `LAST_MODIFIED ASC`" is inaccurate for `assets`, which sorts by `ID`. Same root inaccuracy as the module docstring (collibra.py:30). This is the conceptual error behind the assets data-loss bug; correcting the wording here would help.

src/databricks/labs/community_connector/sources/collibra/connector_spec.yaml
```diff
@@ -0,0 +1,106 @@
...
+        https://databricks.collibra.com/rest/2.0). If provided, overrides
+        org-based URL construction. Useful for on-premises or custom URL
+        routing.
+
+    - name: page_size
+      type: string
+      required: false
+      description: >
+        Page size for paginated API requests. Valid range: 1–1000.
+        Default: 200. Larger pages reduce request count but increase
```
yyoli-db: spec says page_size "Default: 200", but `DEFAULT_PAGE_SIZE = 1000` (collibra_schemas.py:181) and README line 26 says 1000. Reconcile to 1000.

src/databricks/labs/community_connector/sources/collibra/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Troubleshooting
+
+### Authentication errors (401 / 403)
+
+**Symptoms:** `Collibra API error for {table}: 401 ...` or `403 ...`.
+
+**Causes:**
+- The injected bearer token expired or the UC connection could not mint one (bad `client_id` / `client_secret`).
+- The OAuth client lacks the `kg.view-all` scope.
```
yyoli-db: this tells users the OAuth client needs the `kg.view-all` scope, but connector_spec.yaml (lines 90-94) documents as a live-validated finding that Collibra Integration apps reject any `scope` parameter (`invalid_scope`) and must mint with NO scope. This advice would send users down a dead end. Align the README with the spec.

src/databricks/labs/community_connector/sources/collibra/collibra.py
```diff
@@ -0,0 +1,455 @@
...
+            return iter(records), start_offset
+
+        return iter(records), {"cursor": max_seen}
+
+    # ------------------------------------------------------------------ #
+    # Record shaping
+    # ------------------------------------------------------------------ #
+
+    def _shape_common(self, raw: dict[str, Any]) -> dict[str, Any]:
+        """Shape an asset record: normalize nested refs, stamp org."""
```
yyoli-db: this is not for asset only? but a common util?

yyoli-db (changes requested):

wesley84:
Thanks Yong — sharp catches, especially the assets one. Pushed `cc59d2f` addressing all 7.

**The two real bugs:**

- **assets silent loss** — you're right, this was guaranteed at scale. `/assets` sorts by `ID` (`LAST_MODIFIED` 400s), so the count cap advanced the watermark on an arbitrary ID-ordered slice and the next run's `> since` skipped everything at/below it. Went with your **first option**: the assets path now ignores `max_records_per_batch` and drains the full collection per run (bounded by `_init_ts`), so `max_seen` is the true global max.
- **tie-group split** — fixed exactly as you suggested: `max_records` is now a *soft* cap that keeps draining the boundary `lastModifiedOn` group before stopping.

**One thing I want your read on:** I deferred your *second* assets option — the resumable ID-keyset cursor (`id > last_id`) — to a follow-up rather than doing it here. Reasoning: option (a) fully fixes the data loss you filed and is low-risk, whereas the keyset cursor is a multi-field offset state machine (paginate by `id`, filter incrementally by `lastModifiedOn`, only advance the watermark on a *complete* pass) — real logic worth its own PR + tests. It's also the one that'd independently mitigate the m2m token-expiry limitation, since it could resume mid-collection. I've noted it as a planned follow-up in the README + commit msg.

Happy to fold the keyset cursor into this PR instead if you'd rather not merge assets as full-scan-per-run — your call.

Also fixed the 5 doc/naming items (`sort_field` default, `page_size`, the LAST_MODIFIED-ASC wording, the `kg.view-all` scope dead-end, `_shape_common`→`_shape_asset`). Added a regression test per data-loss bug (`test_collibra_incremental_regression.py`); suite green (10 passed / 2 skipped), pylint 10/10.

src/databricks/labs/community_connector/sources/collibra/collibra.py
wesley84: Fixed in `cc59d2f` — the assets path now passes `cursor_sorted=False`, which ignores the count cap and drains the full collection per run so `max_seen` is the true global max. Took your option (a); the ID-keyset cursor (option b) is deferred to a follow-up — see the top-level comment for the reasoning / your call.

src/databricks/labs/community_connector/sources/collibra/collibra.py
wesley84: Fixed — `max_records` is now a soft cap that keeps draining the boundary `lastModifiedOn` tie group before breaking, exactly as you described. Regression test in `test_collibra_incremental_regression.py::test_cursor_sorted_soft_cap_does_not_split_tie_group`.

src/databricks/labs/community_connector/sources/collibra/README.md
wesley84: Fixed — default corrected to `ID` with the 400 caveat noted.

src/databricks/labs/community_connector/sources/collibra/README.md
wesley84: Fixed — reworded: attributes/responsibilities sort `LAST_MODIFIED ASC`, assets sort `ID`. Same correction applied to the module docstring.

src/databricks/labs/community_connector/sources/collibra/connector_spec.yaml
wesley84: Fixed — reconciled to 1000.

src/databricks/labs/community_connector/sources/collibra/README.md
wesley84: Fixed — replaced with 'mint with NO scope' guidance, aligned to the spec's live-validated `invalid_scope` finding.

src/databricks/labs/community_connector/sources/collibra/collibra.py
wesley84: Good call — renamed `_shape_common` -> `_shape_asset` and clarified in the docstring that it's the assets-only default transform (attributes/responsibilities/domains each pass their own shaper).

src/databricks/labs/community_connector/sources/collibra/collibra.py
wesley84: Live-validated on e2-dogfood: ran `assets` scoped to a 790-asset domain with `max_records_per_batch=10` → **790 rows landed, 790 distinct IDs**. The pre-fix code would have truncated at 10 and silently dropped the other ~780; the `cursor_sorted=False` path now drains the full collection despite the cap, with no duplicates (clean resume). Record shaping + m2m no-scope token also confirmed end-to-end.

yyoli-db (approved):

PR #256, opened 2026-07-30
feat(cli): add `publish` command to save a connector as a workspace asset

yyoli-db:
## What

Adds a `publish` subcommand to the community-connector CLI that saves a connector to the user's Databricks workspace as a `.connector.json` asset, at:

```
/Users/<you>/.community-connectors/<display_name>.connector.json
```

The workspace server infers the `CommunityConnector` asset type from the `.connector.json` extension (via `workspace.import_(format=AUTO)`), and the saved connector then surfaces as a **Custom** tile on the Add Data page. This is the CLI counterpart to the webapp's "save community connector to workspace" flow.

## How it works

The manifest mirrors the webapp's own producer shape:

- `id` = `SOURCE_NAME.upper()`, `sourceName`, `displayName`
- `connectionSpec` — read from the connector's `connector_spec.yaml` (resolved the same way as `create_connection --spec`: local source dir, a `--spec` file, a repo URL, or the default repo)
- `logo: null`, empty `repositoryPath` / `readmeUrl` (the workspace UI does not consume these for saved connectors)
- `dependencies` — the connector's `.whl` volume paths

By default the framework + connector wheels are built from the local source and uploaded to a UC Volume (reusing the managed-ingestion `upload` helpers), and their paths are recorded in `dependencies`. Pass `--package` / `--volume-path` to reuse pre-built wheels instead of building.

```bash
# Build wheels from source and publish (display name from the spec / source name)
community-connector publish github

# Custom display name + a pre-built connector wheel (skips building)
community-connector publish github -d "My GitHub" -p ./dist/connector.whl

# Replace an already-published connector at the same path
community-connector publish github -d "My GitHub" --overwrite
```

## Tests

- Display-name resolution (explicit → spec `display_name` → source name; filename-safety validation).
- Manifest field derivation (`logo: null`, dependency inclusion / omission).
- Workspace `mkdirs` + `import_` behavior, including the `--overwrite` collision hint.
- Build-from-source vs `--package` bypass paths.

Full CLI test suite passes (163 in `test_cli.py`); `cli.py` is pylint-clean (10.00/10) under the repo's CI flags. Added a `publish` section to the CLI README.

### DIFF
--- a/tools/community_connector/src/databricks/labs/community_connector_cli/cli.py
+++ b/tools/community_connector/src/databricks/labs/community_connector_cli/cli.py
@@ -3016,5 +3016,98 @@ def publish(
     )
 
 
+def _workspace_object_exists(workspace_client, path: str) -> bool:
+    """Return True if a workspace object exists at ``path``."""
+    try:
+        workspace_client.workspace.get_status(path)
+        return True
+    except Exception as e:
+        if "RESOURCE_DOES_NOT_EXIST" in str(e) or "does not exist" in str(e).lower():
+            return False
+        raise click.ClickException(f"Failed to check {path}: {e}")
+
+
+@main.command("unpublish")
+@click.argument("source_name")
+@click.option(
+    "--display-name",
+    "-d",
+    "display_name",
+    default=None,
+    help="User-facing name of the connector to remove (the <name>.connector.json "
+    "filename). Defaults to the spec's display_name, then the source name — must "
+    "match what was used at publish time.",
+)
+@click.option(
+    "--spec",
+    "-s",
+    "spec_path",
+    default=None,
+    help="Optional: local path to connector_spec.yaml, or a GitHub repo URL. Only "
+    "used to resolve the display name when --display-name is not given.",
+)
+@click.option(
+    "--yes",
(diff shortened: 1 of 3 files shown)


tools/community_connector/src/databricks/labs/community_connector_cli/cli.py
```diff
@@ -2790,5 +2891,136 @@ def upload(
...
+        display_name, connector_spec, source_name
+    )
+    click.echo(f"Display name: {resolved_display_name}")
+
+    workspace_client = _make_workspace_client()
+
+    dest_dir = _resolve_managed_dest_dir(
+        workspace_client, volume_path, catalog, schema, debug
+    )
+    dependencies = _build_and_upload_managed_wheels(
```
yyoli-db: [Nit] Without `--package`, this always calls into `_build_and_upload_managed_wheels(source_name=...)`, which requires the local repo source tree — so `publish <source>` fails outside the repo root with a `Could not find source directory` error. Worth a one-line note in the README/help. (Also the `if not source_name` guard inside that helper is dead on this path since `source_name` is a required arg.)

PR #262, opened 2026-08-02
test(odata): split 18k-line test file into per-feature modules

Leone67:
## Summary

Addresses the review comment on #231 asking to split the gigantic odata test file. `test_odata_lakeflow_connect.py` had grown to ~18,350 lines / 608 module-level tests, which is hard to navigate.

Split into per-feature modules plus a shared helpers module, keeping the CI contract-suite class in the original file.

## What changed

- **`_odata_test_helpers.py`** — the ~40 `*_METADATA_XML` CSDL documents, `SERVICE_URL`, and the `responses` callback/builder helpers, re-exported via `__all__` for `import *`.
- **`test_odata_{contained,cursor,delta,expand,batch,partition,pagination,schema,misc}.py`** — tests grouped by feature prefix.
- **`test_odata_lakeflow_connect.py`** — now just the `TestODataConnector` contract-suite class CI runs (74 lines, down from 18,348), so the self-review B1–B5 greps and CI expectations are unchanged.

| File | Tests |
|---|---|
| test_odata_contained.py | 113 |
| test_odata_cursor.py | 97 |
| test_odata_schema.py | 91 |
| test_odata_misc.py | 79 |
| test_odata_delta.py | 64 |
| test_odata_pagination.py | 63 |
| test_odata_expand.py | 38 |
| test_odata_partition.py | 32 |
| test_odata_batch.py | 31 |

## Verification

Pure mechanical move (AST-driven) — no test logic changed.

- Full suite: **618 passed / 12 skipped** (identical to before the split).
- Each split module also passes standalone.
- Unused per-file imports trimmed.
- Only files under `tests/unit/sources/odata/` are touched.


tests/unit/sources/odata/_odata_test_helpers.py
```diff
@@ -0,0 +1,1393 @@
+"""Shared fixtures, metadata documents, and HTTP helpers for the OData
+connector unit tests.
+
+Extracted from the former monolithic ``test_odata_lakeflow_connect.py`` so the
+per-feature test modules (``test_odata_contained.py`` etc.) can share one copy
+of the ~40 ``*_METADATA_XML`` CSDL documents and the ``responses`` callback
+builders. ``import *`` re-exports every constant and helper via ``__all__``.
```
yyoli-db: **Minor — `import *` is the wrong sharing mechanism here, and the `noqa` on the consuming side is misdirected.**

Three concrete problems with this design:

1. **The `# noqa: F401,F403` on the nine consuming import lines does not suppress what it appears to.** F401/F403 attach to the *import statement*; the **2,335 F405** violations ("may be undefined, or defined from star imports") attach to the *usage* sites and are entirely unsuppressed. Under this repo's own ruff config (`select = ["E","F","I","PLR1730"]`), this PR takes `tests/unit/sources/odata/` from **14 findings to 2,379**.
2. **It's unprecedented in this repo.** No other test file under `tests/` uses `import *` for helpers — qualtrics, hubspot, and example all use explicit imports from a `*_test_utils.py`, and `tests/unit/sources/odata/conftest.py` already exists as the idiomatic fixture seam.
3. **It defeats go-to-definition and static analysis.** A reader of `test_odata_contained.py` can't tell that `_FENCE_MD` or `_make` comes from here without opening this file.

The usual objection — "explicit imports of 72 names would be unwieldy" — doesn't apply: **each module uses only 10–26 of the 72 exported names** (partition 10, batch 12, pagination 13, delta 14, cursor 15, contained 16, expand 18, misc 19, schema 26). That's a 2–4 line explicit import per module:

```python
from tests.unit.sources.odata._odata_test_helpers import (
    NESTED_METADATA_XML, SERVICE_URL, _make, _mock_nested_metadata,
)
```

This drops all 2,335 F405s, restores IDE navigation, matches repo convention, and turns a future stale helper reference into a ruff error instead of a runtime `NameError`.

tests/unit/sources/odata/_odata_test_helpers.py
```diff
@@ -0,0 +1,1393 @@
...
+
+import pytest
+import requests
+import responses
+
+from databricks.labs.community_connector.sources.odata import ODataLakeflowConnect
+from databricks.labs.community_connector.sources.odata.odata import _odata_literal
+from pyspark.sql.types import DecimalType, IntegerType, StringType, TimestampType
+from tests.unit.sources.test_suite import LakeflowConnectTests
+from tests.unit.sources.test_partition_suite import SupportsPartitionedStreamTests
```
yyoli-db: **Minor — 12 of these 13 imports are unused, which contradicts the description's "unused per-file imports trimmed."**

Unused here: `logging`, `os`, `time`, `pytest`, `requests`, `_odata_literal`, `DecimalType`, `IntegerType`, `StringType`, `TimestampType`, `LakeflowConnectTests`, `SupportsPartitionedStreamTests`. Only `json`, `re`, `responses`, and `ODataLakeflowConnect` are actually used — the block looks like it was copied wholesale from the original file.

Note `LakeflowConnectTests`/`SupportsPartitionedStreamTests` are imported *here*, but the class that needs them lives in `test_odata_lakeflow_connect.py`.

Combined with the same issue in `test_odata_lakeflow_connect.py:18-32`, unused imports in this directory go from **3 on master to 25** — the opposite direction from the stated cleanup. Nothing in CI catches it (`pylint.yml:122` excludes `^tests/`, no workflow runs ruff), so this is lint debt rather than breakage. But `.pre-commit-config.yaml` runs `ruff --fix`, so the next contributor with pre-commit installed will silently strip all 25 and carry the noise in an unrelated PR.

Safe to delete now — I verified none of these names are in `__all__`, so no star-importing module can be relying on them. Keeping only `json`, `re`, `responses`, `ODataLakeflowConnect` is exactly what `ruff check --fix --select F401` produces.

tests/unit/sources/odata/_odata_test_helpers.py
```diff
@@ -0,0 +1,1393 @@
+"""Shared fixtures, metadata documents, and HTTP helpers for the OData
+connector unit tests.
+
+Extracted from the former monolithic ``test_odata_lakeflow_connect.py`` so the
+per-feature test modules (``test_odata_contained.py`` etc.) can share one copy
+of the ~40 ``*_METADATA_XML`` CSDL documents and the ``responses`` callback
+builders. ``import *`` re-exports every constant and helper via ``__all__``.
+"""
+
+
```
yyoli-db: Trivial: two blank lines between the module docstring and the first import — `ruff-format` would collapse this to one.

tests/unit/sources/odata/test_odata_delta.py
```diff
@@ -0,0 +1,1762 @@
...
+        )
+        list(records)
+
+
+# ---------------------------------------------------------------------------
+# Round 47 — capability cache LRU cap (memory + disk-inode bound)
+# ---------------------------------------------------------------------------
+
+
+def test_capability_cache_caps_entries_and_sweeps_disk():
```
yyoli-db, line 1609: For the record, this is one of only **two** tests out of 608 whose body is not byte-identical to master — the dropped function-local `import glob`. I confirmed `glob` has **zero** remaining references in this body, so the change is behaviour-preserving and matches the disclosed "unused per-file imports trimmed." No action needed; noting it so it isn't mistaken for an unreviewed edit given that 18k lines moved.

tests/unit/sources/odata/test_odata_delta.py
```diff
@@ -0,0 +1,1762 @@
...
+        for t in threads:
+            t.join()
+    finally:
+        sys.setswitchinterval(prev)
+    assert errors == []
+    assert len(_CAPABILITY_CACHE) == _CAPABILITY_CACHE_MAX_SERVICES
+    _clear_capability_cache()
+
+
+def test_capability_flush_skips_memo_for_evicted_service():
```
yyoli-db, line 1689: The second of the two non-identical bodies: `_CAPABILITY_CACHE` dropped from the function-local import list. Verified **zero** remaining references in this body, so behaviour is unchanged. No action needed.

tests/unit/sources/odata/test_odata_delta.py
```diff
@@ -0,0 +1,1762 @@
...
+        svc = f"https://cap-r47-{i}.example.com/odata/"
+        paths.append(_capability_cache_path(svc))
+        _capability_cache_store(svc, "batch_ok", True)
+    try:
+        # In-memory dict and mtime memo are both bounded at the cap.
+        assert len(_CAPABILITY_CACHE) == _CAPABILITY_CACHE_MAX_SERVICES
+        assert len(_CAPABILITY_DISK_MTIME) <= _CAPABILITY_CACHE_MAX_SERVICES
+        # The most-recent cap services survive; the oldest are evicted.
+        assert f"https://cap-r47-{n - 1}.example.com/odata/" in _CAPABILITY_CACHE
+        assert f"https://cap-r47-0.example.com/odata/" not in _CAPABILITY_CACHE
```
yyoli-db: Trivial / pre-existing (carried over, not introduced here): `F541` f-string without any placeholders — `f"https://cap-r47-0.example.com/odata/"` needs no `f` prefix. Correctly out of scope for a mechanical move; flagging only so it isn't mistaken for a regression, and since you're already touching this file it's a one-character fix.

yyoli-db (changes requested):

Leone67:
Thanks for the thorough review — all addressed in aa75955.

**Star imports → explicit (the main one).** Replaced `from _odata_test_helpers import *` in all 9 split modules with explicit imports of only the names each file uses (AST-computed: partition 10 … schema 26, matching your counts). This eliminates all **2,335 F405s**, restores go-to-definition, and matches the `*_test_utils.py` convention used elsewhere under `tests/`.

**Unused imports.** Trimmed `_odata_test_helpers.py` down to the 4 names its bodies actually use (`json`, `re`, `responses`, `ODataLakeflowConnect` — `time`/`requests` turned out to only appear in docstrings and JSON keys, so they went too), and trimmed `test_odata_lakeflow_connect.py` to what the contract class needs. Also fixed the double blank line after the helpers docstring.

**F541.** Dropped the `f` prefix on the placeholder-less f-string in `test_odata_delta.py`.

Net result: `ruff check` (repo config `E,F,I`) on `tests/unit/sources/odata/` now reports **4 findings, down from 14 on master** — the remaining ones (3 E501 in CSDL XML string literals, 1 E741) are all pre-existing and, as you noted, out of scope for a mechanical move. Full suite still 618 passed / 12 skipped.

yyoli-db (approved):

PR #270, opened 2026-08-20
feat(microsoft_purview): add Microsoft Purview community connector

wesley84:
Adds a read-only **Microsoft Purview** community connector — the Purview adapter for the same "metadata catalog → Unity Catalog" pattern as the merged Collibra connector (#254). Driven by active field demand: a customer wants Purview business metadata hydrated onto UC objects (this connector is the ingest half; hydration is a separate layer).

## Tables (3)
| Table | Endpoint | Ingestion | PK | Cursor |
|---|---|---|---|---|
| `business_domains` | `GET /datagovernance/catalog/businessdomains` | snapshot | `id` | — |
| `data_products` | `GET /datagovernance/catalog/dataProducts` | cdc | `id` | `systemData.lastModifiedAt` |
| `terms` | `GET /datagovernance/catalog/terms` | cdc | `id` | `systemData.lastModifiedAt` |

Targets the **Unified Catalog / Data Governance API** (public preview, api-version `2026-03-20-preview`, base `https://api.purview-service.microsoft.com`).

## Auth
OAuth 2.0 **m2m offloaded to Unity Catalog** (mirrors azure_devops / collibra): UC mints/injects `access_token` from Entra; the connector sends `Authorization: Bearer`, never holds the secret. `token_url = login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token` (CLI `{tenant_id}` interpolation), scope `https://purview.azure.net/.default`. `access_token`/`refresh_token` are not connection params.

## Incremental
Client-side `> since` filter on the ISO-8601 `systemData.lastModifiedAt` cursor + `_init_ts` upper cap for `Trigger.AvailableNow` termination. Standard single-driver read (the Unified Catalog list endpoints have no server-side time filter, so partitioning would full-scan per executor for no benefit).

## Tests / quality
Simulate-mode suite: **7 passed / 2 skipped** (`test_read_terminates` exercises the `_init_ts` cap via `synthesize_future_records`). Pylint **10.00/10** (CI flags). Self-review **90/100 READY** (report posted below).

## Honest caveats — `[needs-live-testing]` (Phase 2, against a live tenant)
- **`max_records_per_batch` not honored on the cdc tables** — the connector full-drains per run because `orderBy` on the nested cursor path isn't verified live. Once `orderBy=systemData/lastModifiedAt asc` is confirmed, a resumable/soft cap can be added (same trajectory as the Collibra assets fix). Deliberately kept out of the allowlist so it's not advertised as a working knob.
- Confirm `nextLink` is absolute + self-contained (embeds `api-version` + skip token) across all three endpoints.
- Soft-delete surfacing (whether deleted products/terms appear with a state flag and whether `lastModifiedAt` advances on delete) — needed before `cdc_with_deletes`.
- Actual 429 / `Retry-After` behavior on a large first-load.

The in-repo `microsoft_purview_api_doc.md` carries a correction banner: an earlier draft claimed the Unified Catalog layer had no REST API — it does (public preview), which is what this connector targets.


wesley84:
# Self-Review — microsoft_purview

**Overall: 90 / 100 — READY**

Run at: 2026-08-19 · Branch `feat/connector-microsoft-purview` · new connector
(Purview Unified Catalog / Data Governance API → UC; Phase 1, needs-live-testing).

> No BLOCKER fails; all tests pass; pylint 10/10; security clean. The E5 finding from the
> first pass (allowlist advertised `max_records_per_batch`, which the code never reads) is
> **fixed** — removed from `external_options_allowlist` and README clarified. One documented
> design caveat remains (A6): the cdc tables full-drain per run because the Unified Catalog
> list endpoints have no verified server-side ordering on the nested `systemData.lastModifiedAt`
> cursor, so `max_records_per_batch` isn't honored yet. Same tradeoff class Collibra's assets
> shipped with pre-fix-(b); the resumable pattern applies once `orderBy` is verified live
> (Phase 2). Plus the expected no-cassette (B11) for a Phase-1 needs-live-testing PR.

## Top recommendations

1. **MAJOR (A6) — no admission control on the cdc tables.** `data_products` / `terms` are
   `cdc` but the connector never reads `max_records_per_batch`; it drains the full
   collection per run (bounded only by the `_init_ts` cap). Deliberate (records aren't in
   verified cursor order, so count truncation can't safely advance the ISO watermark), but
   it means a large tenant does an unbounded single-batch drain — the token-expiry exposure.
   *Phase-2 fix: confirm `orderBy=systemData/lastModifiedAt asc` live, then add a
   page-granular/resumable cap (mirror the collibra assets fix).*
2. *(resolved)* E5 — allowlist advertised `max_records_per_batch` while the code ignored
   it. Fixed this pass: removed from `external_options_allowlist` in `connector_spec.yaml`
   and clarified in the README.
3. **MAJOR-warn (B11)** — no record-mode cassette (Phase 1, no creds). Live validation is
   deferred to Phase 2 against a real tenant (orderBy, nextLink self-containment, soft-delete
   surfacing, 429 behavior). Expected for a needs-live-testing PR.

## A. Connector implementation — 28 / 33
- ✅ A1. Extends `LakeflowConnect` — `microsoft_purview.py:78`
- ✅ A2. Abstract methods present — `list_tables:151, get_table_schema:154, read_table_metadata:160, read_table:166`
- ✅ A3. No `cdc_with_deletes` table → `read_table_deletes` N/A
- ✅ A4. Read pattern deliberate: standard single-driver (Unified Catalog list endpoints have no server-side time filter → partitioning would full-scan per executor; documented in docstring)
- ✅ A5. `_init_ts` cap present + applied — `microsoft_purview.py:129, 286, 297` (ISO-8601 string, lexicographic compare)
- ❌ A6. `max_records_per_batch` NOT read anywhere in the connector — cdc tables full-drain per run. See Top rec #1.
- ✅ A7. All HTTP calls have `timeout=`
- ✅ A8. Schemas use `LongType`/`StructType`, never `IntegerType`/`MapType` — `microsoft_purview_schemas.py:18,21`
- ✅ A9. Nested fields kept as `StructType`, not flattened
- ✅ A10. Imports clean (stdlib / requests / pyspark / interface / intra-source only)
- ✅ A11. Pylint 10.00/10 (CI flag set)
- ✅ A13. `MicrosoftPurviewDataSource(LakeflowSource)` + `_lakeflow_connect_cls` — `__init__.py:9-10`

- ✅ B1–B7. Test file, `LakeflowConnectTests`, connector_class, `simulator_source="microsoft_purview"`, `replay_config` (keys match `__init__`), endpoints.yaml, 3 corpus files
- ✅ B8. Connector URLs resolve against spec (simulate passes)
- ✅ B9. `test_read_terminates` passes
- ✅ B10. Full suite: 7 passed / 2 skipped (skips correct: no cdc_with_deletes; not partitioned)
- ⚠️ B11. No record-mode cassette — Phase 1, live validation deferred to Phase 2 (MAJOR-warn)
- ✅ B13. `synthesize_future_records` present on the 2 cdc endpoints (exercises the `_init_ts` cap)
- ✅ B14. Read-only connector → `test_utils_class` correctly unset

## C. Artifacts — 27 / 27
- ✅ C1. `microsoft_purview.py` compiles
- ✅ C2. `microsoft_purview_api_doc.md` present; covers all 3 tables (+ correction banner on the stale UC-API claim)
- ✅ C3. `connector_spec.yaml` parses; `auth_methods` + `oauth: flow: m2m` (Entra token_url, `{tenant_id}`, scope `.default`); `access_token`/`token` correctly NOT connection params
- ✅ C4. `README.md` present; covers all tables + connection params
- ✅ C5. `pyproject.toml` — requests runtime dep
- ✅ C6. `_generated_microsoft_purview_python_source.py` fresh (mtime ≥ source)
- ✅ C8. 3 corpus files valid JSON
- ✅ C9. Test class present
- ✅ C10. `pyspark` in `[optional-dependencies].dev` only — not runtime

## D. Security — 22 / 22
- ✅ D1–D10. No hardcoded secrets, eval/exec, subprocess, verify=False, unsafe yaml.load, pickle, path traversal, secrets-in-logs, or http://.

- ✅ E1. `list_tables` (business_domains/data_products/terms) ≡ API doc ≡ README
- ✅ E2. Connection params consistent (OAuth exemption: `access_token`/`token` injected, correctly absent from spec params)
- ✅ E4. Primary keys (`id`) consistent
- ✅ E5. `external_options_allowlist` (domain_id, parent_id, keyword, order_by, write_only, page_size) ≡ options the connector reads — `max_records_per_batch` removed this pass (was a no-op knob)
- ✅ E7. README consistent with API doc on auth + pagination (preview-API caveat noted)

wesley84:
**Live-tested ✅** — validated against a live Microsoft Purview (Unified Catalog / Data Governance API) tenant:

- **Direct connector reads** against the live API — all 3 tables (`business_domains`, `data_products`, `terms`): correct record shaping, snapshot vs cdc modes, and cdc cursor → watermark → convergence (re-read from the returned offset yields 0 new records).
- **Full managed pipeline end-to-end** in a Databricks workspace — UC COMMUNITY connection (`service_principal` / m2m; UC mints the Entra token server-side), managed ingestion pipeline **COMPLETED**, data landed in UC Delta tables (`business_domains` 1, `data_products` 1, `terms` 2).

The in-workspace run surfaced — and this branch fixes (`ed5933c`) — a CDC issue: the cdc `cursor_field` was a nested path (`systemData.lastModifiedAt`), which SDP's `APPLY CHANGES` can't use as `sequence_by` (`SEQUENCE_BY_COLUMN_NOT_FOUND`). Fixed by promoting it to a top-level `last_modified_at` column and pointing `cursor_field` at it.

Dropping `[needs-live-testing]` from the title accordingly.

src/databricks/labs/community_connector/sources/microsoft_purview/microsoft_purview_api_doc.md
```diff
@@ -0,0 +1,927 @@
+# Microsoft Purview API Documentation
+
+> ⚠️ **CORRECTION (2026-08-19) — supersedes stale claims below.** An earlier revision of this
+> doc (2026-08-14) stated the Unified Catalog governance layer had **no published REST API**.
+> That is **no longer true** (and was already stale by weeks): Microsoft shipped the **Unified
+> Catalog / Data Governance REST API** — public preview, api-version `2026-03-20-preview` —
+> covering business domains, data products, and terms. Base host
+> `https://api.purview-service.microsoft.com`, paths `/datagovernance/catalog/...`, auth via
+> Entra OAuth2 (scope `https://purview.azure.net/.default`). **This connector targets that
+> Unified Catalog API** (`business_domains`, `data_products`, `terms`). The Atlas Data Map API
```
rahulkhanna-dbl: @wesley84 , can you remove the references to the outdated docs, instead of keeping the version edits on top of it?

src/databricks/labs/community_connector/sources/microsoft_purview/microsoft_purview.py
```diff
@@ -0,0 +1,359 @@
...
+        # Already caught up to init time — nothing new can be emitted.
+        if since is not None and since >= self._init_ts:
+            return iter([]), start_offset
+
+        records: list[dict[str, Any]] = []
+        max_seen = since
+        for raw in record_iter:
+            cursor = self._record_cursor(raw)
+            # Strict `>` so a resumed inclusive `since` doesn't re-emit boundary.
+            if since is not None and cursor is not None and cursor <= since:
```
rahulkhanna-dbl: edge case: since the smallest unit for lastModified is `second`, we can add a comment in here/ docs or take care of this here so that two records with the same second in the lastModifiedat doesn't lose on record

tests/unit/sources/microsoft_purview/test_microsoft_purview_unit.py
rahulkhanna-dbl: can you add a test for  _read_business_domains / _read_data_products / _read_terms methods, where table options are translated into API query params

src/databricks/labs/community_connector/sources/microsoft_purview/microsoft_purview_api_doc.md
wesley84: Done. removed the correction banner and rewrote the stale "Unified Catalog has no REST API" claims throughout (6 spots) to reflect the Data Governance REST API (2026-03-20-preview) the connector actually uses. Data Map is now framed as the alternative surface for classic-Atlas tenants, not the "only" one.

src/databricks/labs/community_connector/sources/microsoft_purview/microsoft_purview.py
wesley84: Good catch. lastModifiedAt is second-granular, so a strict > could permanently drop a same-second record that arrives after the watermark reached that second. Switched to an inclusive >= since boundary (skip only strictly-older rows), boundary rows get re-emitted and are deduped by the CDC primary-key (id) upsert. Bounded to rows at exactly the max second, and the _init_ts cap still guarantees convergence. Updated the docstring/comment and the tests.

tests/unit/sources/microsoft_purview/test_microsoft_purview_unit.py
wesley84: Added TestTableOptionsToQueryParams... asserts each reader translates its options into the right Purview params (domain_id→domainId, parent_id→parentId, keyword, order_by→orderBy, write_only→writeOnly) and omits them when not supplied.

PR #275, opened 2026-09-02
docs(cli): document nested connector_options placement in pipeline specs

wesley84:
## Motivation

When configuring a managed community-connector pipeline, source-specific options (e.g. GitHub's `owner`/`repo`, a `domain_id` filter, `max_records_per_batch`) must be nested under:

```yaml
table:
  connector_options:
    community_connector_options:
      options:
        <source_option>: <value>
```

But the CLI's spec template + generated ingest scripts documented them **flat under `table_configuration`** (the `owner`/`repo` example sat right next to `scd_type`/`primary_keys`). `table_configuration` only carries framework ingestion controls (`scd_type`, `primary_keys`, `sequence_by`); source options placed there are **ignored by the managed flow** and silently never reach the connector — the deployed `ingestion_definition` comes back with them dropped. The misleading examples lead directly to a spec that looks right and quietly does nothing.

The correct nested shape is already documented in the `deploy-connector` skill; this PR brings the CLI's own templates and validator in line with it.

## Changes

- **`templates/pipeline_spec_template.yaml`**, **`templates/ingest_template.py`**, **`templates/ingest_template_base.py`**: show the nested `connector_options.community_connector_options.options` shape and move the GitHub `owner`/`repo` example there; clarify `table_configuration` is ingestion controls only.
- **`pipeline_spec_validator.py`**: add `connector_options` (and `source_schema`, which `managed_pipeline` backfills onto table objects) to `known_table_keys`. Without this, a **correctly-shaped** spec triggers a spurious `Unknown keys in '...table' will be ignored: {'connector_options'}` warning — the validator was contradicting the correct placement. Docstring updated to match.
- **`tests/test_pipeline_spec_validator.py`**: new test asserting a spec with nested `connector_options` produces no warning.

## Verification

- Full CLI test suite: **293 passed** (including the new test).
- YAML template still parses; `ingest_template.py` still parses; `ruff` introduces no new findings (2 pre-existing import-order findings in the test file are untouched).

## Possible follow-up (not in this PR)

The validator could actively **warn** when it sees unrecognized keys flat under `table_configuration` (hint: "did you mean `connector_options.community_connector_options.options`?") — that would catch the mistake at validate time rather than only documenting the right way. Left out here to keep this change doc-focused and low-risk; happy to add it if maintainers want it.


rahulkhanna-dbl (approved):

rahulkhanna-dbl:
@wesley84 can you fix the lint issue:
************* Module tests.test_pipeline_spec_validator
tools/community_connector/tests/test_pipeline_spec_validator.py:373:15: C1803: "warnings == []" can be simplified to "not warnings", if it is strictly a sequence, as an empty list is falsey (use-implicit-booleaness-not-comparison)

wesley84:
> @wesley84 can you fix the lint issue: ************* Module tests.test_pipeline_spec_validator tools/community_connector/tests/test_pipeline_spec_validator.py:373:15: C1803: "warnings == []" can be simplified to "not warnings", if it is strictly a sequence, as an empty list is falsey (use-implicit-booleaness-not-comparison)

@rahulkhanna-dbl Fixed in 0d26b2f — switched to assert not warnings. Pylint's clean now (C1803 10/10) and all 32 tests still pass.

